PR #1, opened 2023-10-06
[Feature/un9] - Adding Dockerfile & Readme

kudlaj:
This first feature contains the following tasks:

- add a simple go project with a json-rpc server that has one method: System.Up
- create the Dockerfile
- create a README


kudlaj:
This first feature contains the following tasks:

- add a simple go project with a json-rpc server that has one method: System.Up
- create the Dockerfile 
- create a README

Dockerfile
```diff
@@ -0,0 +1,32 @@
+# Use an official Go runtime as a parent image
+FROM golang:1.20-alpine AS builder
```
zorjak: can we use the latest version of go?

zorjak (approved):

README.md
```diff
@@ -1 +1,27 @@
...
+## Run your own Node
+
+You can start and sync the universal node locally with the following command:
+```
+$ docker run -p 5001:5001 freeverseio/laos-universal-node:<release> 
+```
+
+The port is for the json-rpc interface.
+
+Right now there is only on method supported:
```
luispdm: Typo: one instead of on

Dockerfile
```diff
@@ -0,0 +1,32 @@
+# Use an official Go runtime as a parent image
```
luispdm: I have a few suggestions for this:
1. choose a fixed version for both builder and final stages. I suggest: `golang:1.21-alpine3.18` and `alpine:3.18.4`
2. change `RUN mkdir...` to `RUN addgroup -g 1000 nodegroup && adduser -D nodeuser -u 1000 -G nodegroup` as it adds clarity
3. remove `go mod download` and `RUN apk --no-cache add ca-certificates` as they are not required
4. you can swap `RUN CGO_ENABLED=1 go build -race -installsuffix 'static' -o app .` with `RUN go build -race -o app .` as we are not linking any C library to our executable at the moment
5. to install the C compiler, you can replace `RUN apk add --no-cache ca-certificates git alpine-sdk linux-headers` with `RUN apk add --no-cache build-base=~0.5`. The `~0.5` is gonna match 0.5 and any minor update made to the package, source: https://wiki.alpinelinux.org/wiki/Alpine_Package_Keeper#Holding_a_specific_package_back. You can see the latest versions of each alpine package here: https://pkgs.alpinelinux.org/packages
6. there's one extra space you can remove between `COPY --from=builder` and `/app`
7. remove all the comments. I think, at least for now, all the instructions are pretty self-explanatory
8. if some instructions take too much space on the line, please use line breaks `\` and indentation

At the end of the day, I managed to end up with the following:
```Dockerfile
FROM golang:1.21-alpine3.18 AS builder
# Add C compiler to build the executable with race detection enabled
RUN apk add --no-cache build-base=~0.5

RUN addgroup -g 1000 nodegroup && \
    adduser -D nodeuser -u 1000 -G nodegroup

WORKDIR /app
COPY ./go .
RUN go build -race -o app .

FROM alpine:3.18.4 AS final

WORKDIR /app
COPY --from=builder /app /app

CMD ["./app"]
USER nobody:nobody

```

Some of the suggestions came from the following tools (lint and vulnerability scanning):
https://trivy.dev/
https://github.com/hadolint/hadolint
https://docs.docker.com/scout/

Clever use of `USER nobody:nobody` 👏

luispdm (changes requested):

go/main.go
```diff
@@ -0,0 +1,57 @@
...
+func main() {
+	s := new(System)
+	rpc.Register(s)
+
+	// Create an HTTP handler for RPC
+	handler := http.NewServeMux()
+	handler.HandleFunc("/rpc", func(w http.ResponseWriter, r *http.Request) {
+		serverCodec := jsonrpc.NewServerCodec(&httpReadWriteCloser{r.Body, w})
+		w.Header().Set("Content-type", "application/json")
+		w.WriteHeader(200)
```
AlexRipoll: I suggest replacing 200 by `http.StatusOK` for greater clarity

go/main.go
```diff
@@ -0,0 +1,57 @@
...
+		err := rpc.ServeRequest(serverCodec)
+		if err != nil {
+			fmt.Println("Error while serving JSON request", err)
```
AlexRipoll: Shouldn't the header status code be updated in the event of an error?

AlexRipoll (changes requested):

AlexRipoll (approved):

PR #6, opened 2023-10-16
feature/UN-10 scan events

luispdm:


internal/scan/scan.go
```diff
@@ -0,0 +1,181 @@
...
+	err := unpackIntoInterface(&approvalForAll, contractAbi, eL)
+	if err != nil {
+		return approvalForAll, err
+	}
+	approvalForAll.Owner = common.HexToAddress(eL.Topics[1].Hex())
+	approvalForAll.Operator = common.HexToAddress(eL.Topics[2].Hex())
+	return approvalForAll, nil
+}
+
+func unpackIntoInterface(e Event, contractAbi *abi.ABI, eL *types.Log) error {
```
kudlaj, line 175: I think it would be better pass eventTransferName as an parameter to the function.
It would be clearer and more robust.

internal/scan/scan_test.go
```diff
@@ -0,0 +1,205 @@
...
+		}).Return(eventLogs, nil)
+
+		events, err := s.ScanEvents(context.Background(), fromBlock, toBlock)
+		if err != nil {
+			t.Fatalf("error occurred when scanning events %v", err.Error())
+		}
+
+		_, ok := events[0].(scan.EventTransfer)
+		if !ok {
+			t.Fatal("error parsing event to EventApproval type")
```
kudlaj, line 57: I think the error msg is wrong here, since those are not EventApproval types

internal/scan/scan_test.go
```diff
@@ -0,0 +1,205 @@
...
+	"go.uber.org/mock/gomock"
+)
+
+const (
+	transferEventHash      = "0xddf252ad1be2c89b69c2b068fc378daa952ba7f163c4a11628f55a4df523b3ef"
+	approveEventHash       = "0x8c5be1e5ebec7d5bd14f71427d1e84f3dd0314c0f7b2291e5b200ac8c7c3b925"
+	approveForAllEventHash = "0x17307eab39ab6107e8899845ad3d59bd9653f200f220920489ca2b5937696c31"
+)
+
+func TestScanEvents(t *testing.T) {
```
kudlaj, line 22: there are no tests for errors. Maybe you should consider adding some?

main.go
```diff
@@ -0,0 +1,200 @@
...
+		}
+		return nil
+	})
+	if err := group.Wait(); err != nil {
+		return err
+	}
+	return nil
+}
+
+func runScan(ctx context.Context, c config.Config, cli scan.EthClient, s scan.Scanner) error {
```
kudlaj, line 128: could you add a test for this?

func TestRunScan(t *testing.T) {
	mockCtrl := gomock.NewController(t)
	defer mockCtrl.Finish()

	mockEthClient := mock.NewMockEthClient(mockCtrl)
	mockScanner := mock.NewMockScanner(mockCtrl)

	mockEthClient.EXPECT().BlockNumber(gomock.Any()).Return(uint64(100), nil).AnyTimes()
	mockScanner.EXPECT().ScanEvents(gomock.Any(), gomock.Any(), gomock.Any()).Return(nil, nil).AnyTimes()

	c := config.Config{
		StartingBlock: 0,
		BlocksRange:   100,
	}

	// Create a context with a timeout
	ctx, cancel := context.WithTimeout(context.Background(), 1*time.Second)
	defer cancel() 
	err := runScan(ctx, c, mockEthClient, mockScanner)
	if err != nil {
		t.Errorf("runScan() returned an error: %v", err)
	}
}

internal/scan/scan.go
```diff
@@ -0,0 +1,181 @@
...
+	Operator common.Address
+	Approved bool
+}
+
+type Scanner interface {
+	ScanEvents(ctx context.Context, fromBlock *big.Int, toBlock *big.Int) ([]Event, error)
+}
+
+type scanner struct {
+	cli      EthClient
```
zorjak, line 74: maybe change name to client. or just c. cli is confusing me a bit because it is common for command line interface:)

internal/scan/scan_test.go
zorjak: I agree with @kudlaj . It would be good to add tests for errors

main.go
```diff
@@ -0,0 +1,200 @@
...
+			if err != nil {
+				slog.Error("error occurred while scanning events", "err", err.Error())
+				break
+			}
+			c.StartingBlock = lastBlock + 1
+		}
+	}
+}
+
+func waitBeforeNextScan(ctx context.Context, waitingTime time.Duration) {
```
zorjak, line 165: can we remove this function from main.go and put it in some other file. Maybe some other package. scan is OK for me.

main.go
```diff
@@ -0,0 +1,200 @@
...
+	timer := time.NewTimer(waitingTime)
+	select {
+	case <-ctx.Done():
+		timer.Stop()
+	case <-timer.C:
+	}
+}
+
+func getL1LatestBlock(ctx context.Context, cli scan.EthClient) (uint64, error) {
```
zorjak, line 174: also place this in some other file. maybe also scan package

main.go
```diff
@@ -0,0 +1,200 @@
...
+	l1LatestBlock, err := cli.BlockNumber(ctx)
+	if err != nil {
+		return 0, err
+	}
+	return l1LatestBlock, nil
+}
+
+func calculateLastBlock(startingBlock, l1LatestBlock uint64, blocksRange, blocksMargin uint) uint64 {
```
zorjak, line 182: also. I think this should be removed from main.

main.go
zorjak, line 128: I think that this function should be part of scan package also. I see `main.go` file where you just start your services. 

Also the function is not so simple. I think we have to have tests for it.

zorjak (changes requested):
some tests are missing that it would be good to write in my opinion.

zorjak (changes requested):
I think that it would be good to add some tests as I mentioned.

luispdm:
closing the PR as the git history has been wrongly rewritten

PR #9, opened 2023-10-17
feature/UN-37 hooks, rebase and best-practices rules

luispdm:


git/pre-commit
```diff
@@ -0,0 +1,80 @@
...
+go mod tidy
+
+#### Check all files for suspicious constructs
+echo "${BGBLUE}>> go vet${NC}"
+go vet ${FILES}
+check_errors $?
+
+#### Launching tests
+echo "${BGBLUE}>> go test -count=1 -race ./...${NC}"
+go test -count=1 -race ./...
```
zorjak, line 68: does this mean that we have to run all the tests before committing? Maybe it is too hard if there are many tests?

luispdm: Yes, that runs all tests. My idea is to run just unit tests in the pre-commit hook (it's all we have for now), so it shouldn't take more than a few seconds. This way you'd know if your changes have broken some tests and you wouldn't have to wait for the CI to tell you

README.md
```diff
@@ -24,4 +24,35 @@ Right now there is only one method supported:
...
+### Best-practices
+Follow best-practices and guidelines published by Google and the golang open-source community, such as:
+
+Effective Go - https://go.dev/doc/effective_go
+
+Go style guide - https://google.github.io/styleguide/go/
+
+Go code review comments - https://github.com/golang/go/wiki/CodeReviewComments
+
+Go test comments - https://github.com/golang/go/wiki/TestComments
```
kudlaj: I think we should discuss this before we add this

git/pre-commit
AlexRipoll: I thinks it's fine to do it this way, if changes to a file breaks some unit test tests, you will need to update them anyway

README.md
luispdm: removed, as agreed

kudlaj (approved):

AlexRipoll (approved):

PR #11, opened 2023-10-24
Feature/un 42

kudlaj:
In this feature we apply the agreed folder structure


internal/scan/scan.go
```diff
@@ -13,7 +13,7 @@ import (
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/core/types"
 	"github.com/ethereum/go-ethereum/crypto"
-	"github.com/freeverseio/laos-universal-node/internal/ERC721"
+	ERC721 "github.com/freeverseio/laos-universal-node/internal/platform/blockchain/erc721"
```
zorjak, line 16: why this? is it some standard not to use lowercase?

zorjak (approved):

AlexRipoll (approved):

deployment/environments/dev/values.yaml
```diff
@@ -0,0 +1,4 @@
+contractAddress: 0xBC4CA0EdA7647A8aB7C2061c2E118A18a936f13D
+rpc: https://eth.llamarpc.com
+debug: false
```
luispdm: Let's enable debug logs in dev

deployment/helmfile.yaml
```diff
@@ -14,3 +24,7 @@ releases:
       - name: universalnode
       - namespace: laos-bigbang
       - imageTag: {{ requiredEnv "TAG" }}
+      - contractAddress: {{ .Values.contractAddress }}
```
luispdm: I'd replace the lines from `contractAddress` to `startingBlock` with `- {{ toYaml .Values | nindent 8 }}`, so we pass the entire map of values to the node helmfile, instead of remapping everytime we add a new value

deployment/charts/universalnode/helmfile.yaml
```diff
@@ -7,3 +7,7 @@ releases:
       values:
         - name: {{ .Values.name }}
         - imageTag: {{ requiredEnv "TAG" }}
+        - contractAddress: {{ .Values.name  }}
```
luispdm: Similarly to the parent helmfile, use `- {{ toYaml .Values | nindent 8 }}`, but this time you can replace all the parameters (also name and imageTag) as they are already coming from the parent helmfile's map

deployment/environments/dev/values.yaml
```diff
@@ -0,0 +1,4 @@
...
+startingBlock: 41375688
```
luispdm: If you follow my suggestions of both helmfile.yaml, replace 41375688 with "41375688", otherwise the value will be passed as: 4.1375688e+07

luispdm (changes requested):

luispdm (approved):

PR #12, opened 2023-10-26
Feature/un 14  Json-RPC proxy

kudlaj:
In this feature we create a Json-RPC proxy. 
All Json-RPC to our node will be forwarded to the underlaying BC node.

Now we can connect with Metamask to the universal-node and operate with it as if we would with a BC node directly.


go.mod
```diff
@@ -20,6 +20,7 @@ require (
 	github.com/go-ole/go-ole v1.2.5 // indirect
 	github.com/go-stack/stack v1.8.1 // indirect
 	github.com/google/uuid v1.3.0 // indirect
+	github.com/gorilla/mux v1.8.0 // indirect
```
luispdm: This is a direct dependency. Run `go mod tidy` and commit the change please

cmd/main.go
```diff
@@ -74,51 +46,17 @@ func run() error {
...
-		}
-		return nil
-	})
-	group.Go(func() error {
-		sys := new(System)
-		err := rpc.Register(sys)
+		rpcServer, err := server.NewServer()
 		if err != nil {
-			return fmt.Errorf("error registering RPC service: %w", err)
+			slog.Error("failed to create RPC server: %v", err)
```
luispdm: No need to log the error, return it because `main()` will log it for you:
```go
return fmt.Errorf("failed to create RPC server: %w", err)
```

cmd/main.go
```diff
@@ -74,51 +46,17 @@ func run() error {
...
+			return err
 		}
-		if srvErr := server.ListenAndServe(); srvErr != nil && srvErr != http.ErrServerClosed {
-			return srvErr
-		}
-		return nil
+		addr := fmt.Sprintf("0.0.0.0:%v", c.Port)
+		slog.Info("starting RPC server", "listenAddress", addr)
```
luispdm: `listen_address` instead of `listenAddress` when logging fields names with slog.
I just noticed we wrote `eventType` instead of `event_type` [here](https://github.com/freeverseio/laos-universal-node/blob/main/internal/scan/scan.go#L132), can you change that too, please? Thanks

cmd/server/server.go
```diff
@@ -0,0 +1,116 @@
...
+
+// WithHTTPServer allows you to provide a custom HTTPServerer implementation.
+func WithHTTPServer(httpServer HTTPServerController) ServerOption {
+	return func(s *Server) error {
+		s.HTTPServer = httpServer
+		return nil
+	}
+}
+
+func NewServer(opts ...ServerOption) (*Server, error) {
```
luispdm: I still prefer `server.New()` 😆. Even more so because of `server.New(server.WithHTTPServer())`

cmd/server/server.go
```diff
@@ -0,0 +1,116 @@
...
+
+		s.HTTPServer.SetKeepAlivesEnabled(false)
+		if err := s.HTTPServer.Shutdown(ctx); err != nil {
+			// Error from closing listeners, or context timeout:
+			slog.Error("HTTP server Shutdown: %w", err)
+		}
+	}()
+
+	if err := s.HTTPServer.ListenAndServe(); err != http.ErrServerClosed {
+		slog.Error("HTTP server ListenAndServe: %w", err)
```
luispdm: Same as in `main`: you don't need to log the error here as it is logged by `main()`

cmd/server/server.go
```diff
@@ -0,0 +1,116 @@
...
+		<-ctx.Done()
+		// We received an interrupt signal, shut down.
+		slog.Info("received server shutdown signal. Shutting down gracefully...")
+		ctx, cancel := context.WithTimeout(context.Background(), 15*time.Second)
+		defer cancel()
+
+		s.HTTPServer.SetKeepAlivesEnabled(false)
+		if err := s.HTTPServer.Shutdown(ctx); err != nil {
+			// Error from closing listeners, or context timeout:
+			slog.Error("HTTP server Shutdown: %w", err)
```
luispdm: When using slog, log the field name too, not just the error, like: `slog.Error("HTTP server shutdown", "err", err.Error())`

cmd/server/server.go
```diff
@@ -0,0 +1,116 @@
...
+
+// ListenAndServe starts the RPC server to listen and serve incoming requests on the specified address.
+// It also handles graceful shutdown on receiving a context cancellation signal.
+func (s Server) ListenAndServe(ctx context.Context, rpcUrl, addr string) error {
+	s.HTTPServer.SetAddr(addr)
+
+	handler := api.NewApiHandler(rpcUrl)
+	router := mux.NewRouter()
+	s.HTTPServer.SetHandler(api.NewRouter(handler, router))
+	slog.Info(
```
luispdm: I understand the usage of `slog.String` but for some reason `slog` fails to print `source` in this instance. I see this in my terminal:
```
time=2023-10-27T17:06:29.447+02:00 level=INFO source=~/laos-universal-node/cmd/main.go:56 msg="starting RPC server" version=undefined listenAddress=0.0.0.0:5001
time=2023-10-27T17:06:29.447+02:00 level=INFO source=:0 msg="INFO server listening address=0.0.0.0:5001" version=undefined
```
I got `source=:0` on the second line. If you see the same happening to you, can you change the line to: `slog.Info("server listening", "address", addr)` please?

cmd/server/server.go
```diff
@@ -0,0 +1,116 @@
...
+func (h *HTTPServer) SetAddr(addr string) {
+	h.server.Addr = addr
+}
+
+func (h *HTTPServer) SetHandler(handler http.Handler) {
+	h.server.Handler = handler
+}
+
+type Server struct {
+	HTTPServer HTTPServerController
```
luispdm: Does this variable need to be exported?

cmd/server/api/handler.go
```diff
@@ -0,0 +1,54 @@
+package api
+
+import (
+	"net/http"
+	"time"
+)
+
+// Define an interface for HTTP client operations
+type HttpClientInterface interface {
```
luispdm: For sake of consistency, rename Http to HTTP, please

cmd/server/api/handler.go
```diff
@@ -0,0 +1,54 @@
...
+	Do(req *http.Request) (*http.Response, error)
+}
+
+// Create a wrapper for the net/http client
+type HttpClientWrapper struct {
+	Client *http.Client
```
luispdm: `Client` can be unexported

cmd/server/api/handler.go
```diff
@@ -0,0 +1,54 @@
...
+func (h *HttpClientWrapper) Do(req *http.Request) (*http.Response, error) {
+	return h.Client.Do(req)
+}
+
+type ApiHandlerInterface interface {
+	PostRpcHandler(w http.ResponseWriter, r *http.Request)
+}
+
+type ApiHandler struct {
+	RpcUrl     string
```
luispdm: I see both `RpcUrl` and `HttpClient` are used outside of the package just for testing purposes. Do you want to add getter methods to retrieve the fields and make them unexported?

cmd/server/api/handler_test.go
```diff
@@ -0,0 +1,31 @@
...
+
+	mockHttpClient := mock.NewMockHttpClientInterface(ctrl)
+	rpcUrl := "https://polygon-mumbai.test.com/"
+	handler := api.NewApiHandler(
+		rpcUrl,
+		api.WithHttpClient(mockHttpClient),
+	)
+
+	if handler == nil {
+		t.Fatal("handler is nil")
```
luispdm: I suggest replacing this and the next `Fatal` for `Error`. so if multiple checks fail, we will see all the errors on the test output. You can leave the last one as `Fatalf` as it's the last check in the test

cmd/server/api/handler_test.go
```diff
@@ -0,0 +1,31 @@
...
+	}
+	if handler.HttpClient == nil {
+		t.Fatal("handler.HttpClient is nil")
+	}
+	if handler.RpcUrl != rpcUrl {
+		t.Fatalf("RpcUrl expected %v, got %v", rpcUrl, handler.RpcUrl)
```
luispdm: Please use the syntax `got ... want/expected...` in all of the tests of the PR

cmd/server/api/rpchandler_test.go
```diff
@@ -0,0 +1,134 @@
...
+	}{
+		{
+			name:           "successful request",
+			requestBody:    `{"jsonrpc":"2.0","method":"net_version","params":[],"id":67}`,
+			mockResponse:   `{"jsonrpc":"2.0","result":"1001","id":67}`,
+			expectedStatus: http.StatusOK,
+			expectedBody:   `{"jsonrpc":"2.0","result":"1001","id":67}`,
+		},
+		{
+			name: "successful eth_call request with parrams",
```
luispdm: Typo, `parrams`

cmd/server/api/rpchandler.go
```diff
@@ -0,0 +1,70 @@
+package api
+
+import (
+	"bytes"
+	"io"
+	"log/slog"
+	"net/http"
+)
+
+func (h *ApiHandler) PostRpcHandler(w http.ResponseWriter, r *http.Request) {
```
luispdm: For sake of consistency, I'd also rename Rpc to `RPC`

cmd/server/api/rpchandler.go
```diff
@@ -0,0 +1,70 @@
...
+	// Read the body of the incoming request
+	body, err := io.ReadAll(r.Body)
+	if err != nil {
+		http.Error(w, "Bad Request", http.StatusBadRequest)
+		return
+	}
+	defer func() {
+		errClose := r.Body.Close()
+		if errClose != nil {
+			slog.Error("Error closing response body", "error", errClose)
```
luispdm: Start log messages with lowercase letters and rename all the fields' keys "error" to "err" in all of the file, please

cmd/server/api/rpchandler.go
```diff
@@ -0,0 +1,70 @@
...
+
+	// Prepare the request to the BC node
+	proxyReq, err := http.NewRequest(r.Method, h.RpcUrl, io.NopCloser(bytes.NewReader(body)))
+	if err != nil {
+		http.Error(w, "Internal Server Error", http.StatusInternalServerError)
+		return
+	}
+
+	// Forward headers the request
+	for name, values := range r.Header {
```
luispdm, line 32: Curiosity time: what happens if you replace this block of code with `proxyReq.Header = r.Header`? Will it work?

cmd/server/api/rpchandler_test.go
```diff
@@ -0,0 +1,134 @@
...
+				t.Errorf("expected status %v, got %v", tt.expectedStatus, response.StatusCode)
+			}
+			if string(body) != tt.expectedBody {
+				t.Errorf("expected body %v, got %v", tt.expectedBody, string(body))
+			}
+		})
+	}
+}
+
+func gzipString(s string) io.ReadCloser {
```
luispdm: If you wish, you can modify the signature of the function to `gzip(t *testing.T, s string) io.ReadCloser` and as a first instruction write [`t.Helper()`](https://pkg.go.dev/testing#T.Helper). This way you can call `t.Fatal` instead of `panic`

cmd/server/api/rpchandler_test.go
```diff
@@ -0,0 +1,134 @@
...
+			requestBody:    `{"jsonrpc":"2.0","method":"net_version","params":[],"id":67}`,
+			mockError:      errors.New("client error"),
+			expectedStatus: http.StatusBadGateway,
+			expectedBody:   "Bad Gateway\n",
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			request := httptest.NewRequest(http.MethodPost, "/rpc", bytes.NewBufferString(tt.requestBody))
```
luispdm, line 91: I suggest adding headers to the request and to the mocked response coming from the "BC node", so the test can validate that the headers are properly forwarded in both steps

cmd/server/api/rpchandler_test.go
```diff
@@ -0,0 +1,134 @@
...
+			}
+
+			handler.PostRpcHandler(recorder, request)
+
+			response := recorder.Result()
+			body, _ := io.ReadAll(response.Body)
+			defer func() {
+				errClose := response.Body.Close()
+				if errClose != nil {
+					t.Errorf("Error closing response body: %v", errClose)
```
luispdm: It's usually a convention to do `t.Fatal` when writing table-tests. As you wish, I don't have a strong will to change it, I understand that `t.Error` makes sense here as there are multiple checks

cmd/server/api/router.go
```diff
@@ -0,0 +1,17 @@
...
+
+	"github.com/gorilla/mux"
+)
+
+type Router interface {
+	HandleFunc(pattern string, handler func(http.ResponseWriter, *http.Request)) *mux.Route
+	ServeHTTP(w http.ResponseWriter, r *http.Request)
+}
+
+func NewRouter(h ApiHandlerInterface, r Router) Router {
```
luispdm: I understand the purpose of this function, but to me it is counterintuitive calling it `NewRouter`. When the caller uses this function, they are expected to pass in the router that is later returned.
I would call it `RegisterRoute(s)/SetRoute(s)` but I am open to other options. What do you think?

cmd/server/api/handler.go
```diff
@@ -0,0 +1,54 @@
...
+// Create a wrapper for the net/http client
+type HttpClientWrapper struct {
+	Client *http.Client
+}
+
+func (h *HttpClientWrapper) Do(req *http.Request) (*http.Response, error) {
+	return h.Client.Do(req)
+}
+
+type ApiHandlerInterface interface {
```
AlexRipoll: I suggest removing Api prefix in the structs defined in this package since the package is already named api.

cmd/server/api/handler.go
```diff
@@ -0,0 +1,54 @@
...
+
+type ApiHandlerOption func(*ApiHandler)
+
+func WithHttpClient(client HttpClientInterface) ApiHandlerOption {
+	return func(h *ApiHandler) {
+		h.HttpClient = client
+	}
+}
+
+func NewApiHandler(rpcUrl string, opts ...ApiHandlerOption) *ApiHandler {
```
AlexRipoll: Similar to the previous comment. You should avoid repeating the package name for naming structs, funcs, interfaces... to make it idiomatic as the official Go guide explains it here: https://go.dev/doc/effective_go#names

cmd/server/api/handler.go
```diff
@@ -0,0 +1,54 @@
...
+}
+
+type ApiHandler struct {
+	RpcUrl     string
+	HttpClient HttpClientInterface // Inject the HTTP client interface here
+}
+
+type ApiHandlerOption func(*ApiHandler)
+
+func WithHttpClient(client HttpClientInterface) ApiHandlerOption {
```
zorjak: I don't really understand the need for this. Do we have an  idea for ApiHandler to work without httpClient?Where? If not why do we need explicit functions for this?

cmd/server/api/rpchandler_test.go
```diff
@@ -0,0 +1,134 @@
...
+			name:           "client error",
+			requestBody:    `{"jsonrpc":"2.0","method":"net_version","params":[],"id":67}`,
+			mockError:      errors.New("client error"),
+			expectedStatus: http.StatusBadGateway,
+			expectedBody:   "Bad Gateway\n",
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
```
AlexRipoll, line 79: You can add t.Parallel() to run each test case in parallel

AlexRipoll (changes requested):

luispdm (changes requested):

cmd/server/api/rpchandler_test.go
```diff
@@ -0,0 +1,134 @@
...
+			recorder := httptest.NewRecorder()
+
+			if tt.mockError != nil {
+				mockHttpClient.EXPECT().Do(gomock.Any()).Return(nil, tt.mockError).Times(1)
+			} else {
+				mockResponse := &http.Response{
+					StatusCode: http.StatusOK,
+					Body:       io.NopCloser(strings.NewReader(tt.mockResponse)),
+				}
+				if tt.responseGzip {
```
zorjak: shoudl we still have this? I think we resolved Gzip thing right?

cmd/server/api/rpchandler_test.go
```diff
@@ -0,0 +1,134 @@
...
+			requestBody:    `{"jsonrpc":"2.0","method":"net_version","params":[],"id":67}`,
+			mockResponse:   `{"jsonrpc":"2.0","error":{"code":-32601,"message":"The method net_version does not exist/is not available"}}`,
+			expectedStatus: http.StatusOK,
+			expectedBody:   `{"jsonrpc":"2.0","error":{"code":-32601,"message":"The method net_version does not exist/is not available"}}`,
+		},
+		{
+			name:           "non successful request with invalid json",
+			requestBody:    `{"jsonrpc":"2.0","method":"net_version","params":[],"id":67`,
+			mockResponse:   `{"jsonrpc":"2.0","error":{"code":-32700,"message":"Parse error"}}`,
+			expectedStatus: http.StatusOK,
```
zorjak, line 65: this should return http.StatusOK even if error happens? I am just asking:)

cmd/server/api/rpchandler_test.go
zorjak: do we still need to worry about this gzip? I think we agreed that the node is going to transparent to all HTTP calls. and gzip is nothing special

cmd/server/api/rpchandler_test.go
luispdm: > You can add t.Parallel() to run each test case in parallel

If you go this way, write `t.Parallel()` in both the test and the sub-test

zorjak (approved):

zorjak (changes requested):
some comments left. lets resolve it before I approve this

cmd/server/api/router.go
AlexRipoll: I agree, maybe naming it `Router`, `Routes` or `RegisteredRoutes` would be more suitable. However, I would refrain from naming it 'SetRoutes' since the function itself doesn't actually set any routes; it merely manages routes associated with a function.

cmd/server/api/rpchandler_test.go
kudlaj: yes

cmd/server/api/rpchandler_test.go
```diff
@@ -0,0 +1,142 @@
...
+			t.Cleanup(func() {
+				ctrl.Finish()
+			})
+
+			mockHttpClient := mock.NewMockHttpClientInterface(ctrl)
+			handler := api.NewHandler(
+				"https://polygon-mumbai.g.alchemy.com/",
+				api.WithHttpClient(mockHttpClient),
+			)
+			t.Parallel() // Run tests in parallel
```
luispdm: I have two more requests: move this (`t.Parallel()`) as a first instruction of the sub-test and add `t.Parallel()` to the rest of the tests you've added in this PR, please

luispdm (changes requested):

luispdm (approved):

AlexRipoll (approved):

zorjak (approved):

PR #13, opened 2023-10-27
feature/UN-46 - scan new universal events

luispdm:
- contracts with their related infos are stored in a text file for now
- scanning of erc721 universal events and other events (transfer, approval, approval for all) is sequential for now
- the app now accepts a list of contracts in input: if the user provides it, the app will only search for information on those contracts


internal/scan/store.go
```diff
@@ -0,0 +1,104 @@
...
+	err = os.Chmod(filename, 0o600)
+	if err != nil {
+		return nil, err
+	}
+
+	return fsStorage{file: filename}, nil
+}
+
+// Store adds an ERC721BridgelessContract struct in JSON format to the storage file
+func (fs fsStorage) Store(ctx context.Context, c ERC721BridgelessContract) error {
```
kudlaj: you're not using context, please remove it

internal/scan/store.go
```diff
@@ -0,0 +1,104 @@
...
+	buf = append(buf, '\n')
+	if _, err := f.Write(buf); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// ReadAll implements FSStorage.
+func (fs fsStorage) ReadAll(ctx context.Context) ([]ERC721BridgelessContract, error) {
```
kudlaj: same as above: you're not using context, please remove it

internal/scan/scan_test.go
```diff
@@ -229,25 +229,338 @@ func TestScanEvents(t *testing.T) {
...
+				storage.EXPECT().ReadAll(context.Background()).Return(contracts, nil).Times(1)
+				cli.EXPECT().FilterLogs(context.Background(), ethereum.FilterQuery{
+					FromBlock: fromBlock,
+					ToBlock:   toBlock,
+					Addresses: []common.Address{address},
+				}).Return(nil, tt.error)
+
+				_, err := s.ScanEvents(context.Background(), fromBlock, toBlock)
+				if err == nil {
+					t.Fatal("error expected, got nil")
```
kudlaj: please adjust to the got / expected pattern for all checks

internal/scan/scan.go
```diff
@@ -13,19 +13,22 @@ import (
...
+	eventApprovalForAllName                = "ApprovalForAll"
+	eventNewERC721BridgelessMintingName    = "NewERC721BridgelessMinting"
+	eventTransferSig                       = []byte(fmt.Sprintf("%s(address,address,uint256)", eventTransferName))
+	eventApprovalSig                       = []byte(fmt.Sprintf("%s(address,address,uint256)", eventApprovalName))
+	eventApprovalForAllSig                 = []byte(fmt.Sprintf("%s(address,address,bool)", eventApprovalForAllName))
+	eventNewERC721BridgelessMintingSig     = []byte(fmt.Sprintf("%s(address,string)", eventNewERC721BridgelessMintingName))
+	eventTransferSigHash                   = crypto.Keccak256Hash(eventTransferSig).Hex()
+	eventApprovalSigHash                   = crypto.Keccak256Hash(eventApprovalSig).Hex()
+	eventApprovalForAllSigHash             = crypto.Keccak256Hash(eventApprovalForAllSig).Hex()
+	eventNewERC721BridgelessMintingSigHash = crypto.Keccak256Hash(eventNewERC721BridgelessMintingSig).Hex()
```
kudlaj: since this is growing I would suggest a map[string]string to organise the signatures

internal/scan/scan.go
```diff
@@ -70,27 +73,129 @@ type EventApprovalForAll struct {
...
+		storage: s,
 	}
+	for _, c := range contracts {
+		scan.contracts = append(scan.contracts, common.HexToAddress(c))
+	}
+	return scan
+}
+
+// ScanEvents returns the ERC721 events between fromBlock and toBlock
+func (s scanner) ScanNewBridgelessMintingEvents(ctx context.Context, fromBlock, toBlock *big.Int) error {
```
kudlaj: Please consider to break this method down into smaller functions for better readability

internal/scan/scan.go
```diff
@@ -70,27 +73,129 @@ type EventApprovalForAll struct {
...
+			}
+		default:
+			slog.Debug("no new bridgeless minting contracts found")
+		}
+	}
+
+	return nil
+}
+
+func (s scanner) triggerDiscovery(ctx context.Context) (bool, error) {
```
kudlaj: this name is misleading since it is no triggering the discovery. I would suggest to give it a more accurate name

kudlaj (changes requested):

cmd/main.go
```diff
@@ -145,19 +149,25 @@ func runScan(ctx context.Context, c config.Config, client scan.EthClient, s scan
 				slog.Error("error retrieving the latest block", "err", err.Error())
 				break
 			}
-			lastBlock := calculateLastBlock(c.StartingBlock, l1LatestBlock, c.BlocksRange, c.BlocksMargin)
-			if lastBlock < c.StartingBlock {
+			lastBlock := calculateLastBlock(startingBlock, l1LatestBlock, c.BlocksRange, c.BlocksMargin)
+			if lastBlock < startingBlock {
```
zorjak, line 90: can we put this check inside the function `calculateLastBlock`?

internal/platform/blockchain/ERC721BridgelessMinting/ERC721BridgelessMinting.go
```diff
@@ -0,0 +1,1282 @@
+// Code generated - DO NOT EDIT.
```
zorjak: maybe this file we should placed in a folder `autogenerated`? And put readme how it is autogenerated. just 2 3 lines readme. abigen command. Just a suggestion

internal/scan/scan.go
```diff
@@ -13,19 +13,22 @@ import (
...
-	eventTransferSig           = []byte(fmt.Sprintf("%s(address,address,uint256)", eventTransferName))
-	eventApprovalSig           = []byte(fmt.Sprintf("%s(address,address,uint256)", eventApprovalName))
-	eventApprovalForAllSig     = []byte(fmt.Sprintf("%s(address,address,bool)", eventApprovalForAllName))
-	eventTransferSigHash       = crypto.Keccak256Hash(eventTransferSig).Hex()
-	eventApprovalSigHash       = crypto.Keccak256Hash(eventApprovalSig).Hex()
-	eventApprovalForAllSigHash = crypto.Keccak256Hash(eventApprovalForAllSig).Hex()
+	eventTransferName                      = "Transfer"
+	eventApprovalName                      = "Approval"
+	eventApprovalForAllName                = "ApprovalForAll"
+	eventNewERC721BridgelessMintingName    = "NewERC721BridgelessMinting"
```
zorjak: so when we use new version of contract we have to change these names right? does it make sense to change that in this PR?

internal/scan/scan.go
zorjak: maybe this function should return array of events. And another function that deals with saving events should save them (either in db or file or not save them at all).

zorjak (changes requested):

internal/scan/scan.go
AlexRipoll: These names have already been updated, you may be seeing an outdated version of the PR. Please pull the changes to see the updated version

cmd/main.go
AlexRipoll: This logic stems from a previous PR, wherein we encountered the same issue and made a deliberate choice for the following reason: If we were to relocate this logic from its current function, we would lose the ability to include the 'slog.Debug' component. While it's possible to handle logging in a different function, the ultimate function responsible for recording any form of information should remain the same.

internal/scan/scan.go
AlexRipoll: We've cut down on the number of variables with the new function in the latest commit. Right now, this should do the trick because we're not thinking about scanning more events. But if things change down the road, we can chat about a different approach.

internal/scan/scan.go
AlexRipoll: To be honest there is not much more we can do as we have already broken it down into multiple functions. If you like we can discuss another approach together.

internal/platform/blockchain/ERC721BridgelessMinting/ERC721BridgelessMinting.go
luispdm: We decided to structure it this way:
```
internal
  platform
    blockchain
      contract
        erc721universal.go
```
The motivation was that if for every autogenerated file we have to have a package called `autogenerated`, then we'd have the same situation with mocks, and personally I don't like a mock to belong to a package called `autogenerated` as it is too broad.
Does that work for you?

We have not written any instructions as for now there is no need for more contracts

cmd/main.go
zorjak: To be honest, I would place this check inside the function. Also, I don't understand why you cannot place the log inside the function. But I won't push this if you want to merge PR and finish your task

kudlaj (approved):

internal/platform/blockchain/ERC721BridgelessMinting/ERC721BridgelessMinting.go
zorjak: yes, the folder name `contract` is OK. I like it.

zorjak (approved):

cmd/main.go
luispdm: We decided to create a function to calculate the last block as it was too verbose to show it in the caller code.
We decided not to do the same with `lastBlock < startingBlock` as it's just a simple comparison and it's easy to understand for a reader.

We could move the comparison within `calculateLastBlock`, but then the method would do three things: calculate the last block, compare it with the starting block, and return it. Finding a name for that function won't be easy, and, in my personal opinion, it would be counterintuitive for a reader to look at a method that both calculates and compares, while also returning a piece of data.

Regarding the `slog.Debug` call: we think it's the responsibility of the caller to do that as it knows the flow of execution, the underlying function doesn't have to know that when the last block is behind the starting block, the caller wants to wait.

PR #15, opened 2023-11-03
Feature/un 40 Public Node Deployment

kudlaj:
In this feature we add the deployment for 2 public nodes:

- One on top of polygon-mumbai
- One on top of polygon-mainnet


deployment/environments/dev/values.yaml
```diff
@@ -1,6 +1,18 @@
-contracts: 0xD0b57307a4B9e0147F389ab99547E1530D323be0
-rpc: https://polygon-mumbai-bor.publicnode.com
-debug: true
-startingBlock: "41924130"
-blocksMargin: 70
-blocksRange: 2
+networks:
+  - name: polygon-mumbai
+    contracts: 0xD0b57307a4B9e0147F389ab99547E1530D323be0
+    rpc: https://polygon-mumbai-bor.publicnode.com
```
zorjak: can this be in secrets (or config) so that we can change L1 if this L1 does not work?

deployment/environments/dev/values.yaml
```diff
@@ -1,6 +1,18 @@
...
+    host: polygon-mumbai.unode.gorengine.com
+    debug: true
+    startingBlock: "41924130"
+    blocksMargin: 70
+    blocksRange: 2
+  - name: polygon-mainnet
+    contracts: 0xBC4CA0EdA7647A8aB7C2061c2E118A18a936f13D
+    rpc: https://eth.llamarpc.com
```
zorjak: same as previous comment. Also, is this really polygon mainnet rpc?

zorjak (changes requested):

deployment/charts/universalnode/templates/service.yaml
```diff
@@ -9,5 +9,5 @@ spec:
     app.kubernetes.io/app: {{ .Values.name }}
   ports:
   - name: rpc
-    port: 5001
-    targetPort: rpc
+    port: 80
+    targetPort: 5001
```
luispdm: As targetPort you can leave `rpc`

deployment/environments/dev/values.yaml
```diff
@@ -1,6 +1,18 @@
...
+  - name: polygon-mumbai
+    contracts: 0xD0b57307a4B9e0147F389ab99547E1530D323be0
+    rpc: https://polygon-mumbai-bor.publicnode.com
+    host: polygon-mumbai.unode.gorengine.com
+    debug: true
+    startingBlock: "41924130"
+    blocksMargin: 70
+    blocksRange: 2
+  - name: polygon-mainnet
+    contracts: 0xBC4CA0EdA7647A8aB7C2061c2E118A18a936f13D
```
luispdm: Either we set an already existing contract in polygon mainnet (that one was deployed on ethereum mainnet) or we remove this value.

If we remove contracts entirely, I suggest setting startingBlock to `0`: it is better that the app starts filtering events of the ownership chain from the current block so it does not get rate limited from the L1 node (when startingBlock is 0 the app will actually fetch the latest block and start scanning from there, to avoid scanning a ton of events from "the beginning of times").

luispdm (changes requested):
Please increase the Chart version to `0.2.0`

deployment/environments/dev/values.yaml
kudlaj: since we're using a public node it does not make much sense to put it in the secrets. If you want to change it, it's easier to change it in the deployment

deployment/environments/dev/values.yaml
kudlaj: same

deployment/environments/dev/values.yaml
zorjak: I don't understand why it is different if we use public nodes. The point is that we want to be able to change L1 (if we need to do it) without the need to modify the helm chart, modify git code etc.

zorjak (approved):

deployment/environments/dev/values.yaml
kudlaj: but you can change it without modifying code... it's the same only in the secret you have it encrypted

deployment/environments/dev/values.yaml
luispdm: This endpoint must be changed for `https://polygon-bor.publicnode.com`

luispdm: We must change the values of both `rpc`.
Running the unode with that L1 mumbai node, the following error occurs:
`error filtering events: Please specify an address in your request or, to remove restrictions, order a dedicated full node here: https://www.allnodes.com/matic/host`

and the mainnet value refers to an L1 node of Ehtereum mainnet.

I suggest we use https://rpc-mumbai.maticvigil.com for mumbai and https://polygon.llamarpc.com for mainnet.
If you wish to check other public nodes, you can check this page: https://chainlist.org/?testnets=true&search=polygon

luispdm (approved):

PR #17, opened 2023-11-07
Feature/un 38 detect ERC721 call for universal minting and return an error for now

kudlaj:
In this feature, we detect all those JSON-RPC calls to ERC721 methods that will be covered by the Universal node and are not yet implemented.
We return an error until those methods are implemented.

The methods are:

- ownerOf()
- balanceOf()
- totalSupply()
- tokenOfOwnerByIndex(owner, index)
- tokenByIndex(index)


cmd/server/api/rpcmiddleware.go
```diff
@@ -0,0 +1,144 @@
...
+		erc721Handler.ServeHTTP(w, r)
+		return
+	} else {
+		standardHandler.ServeHTTP(w, r)
+		return
+	}
+}
+
+// reportError is a utility function for reporting errors through HTTP.
+func reportError(w http.ResponseWriter, message string, statusCode int) {
```
AlexRipoll: What is the purpose of this function? In my opinion, it doesn't  add any value

cmd/server/api/router.go
```diff
@@ -11,7 +12,12 @@ type Router interface {
...
-func Routes(h HandlerInterface, r Router) Router {
-	r.HandleFunc("/", h.PostRPCHandler).Methods("POST")
-	return r
+func Routes(h HandlerInterface, r Router, storage scan.Storage) Router {
+	router := r.(*mux.Router)
+	rpcProxyHandler := http.HandlerFunc(h.PostRPCProxyHandler)
+	erc721UniversalMintingHandler := http.HandlerFunc(h.UniversalMintingRPCHandler)
+
+	// Pass both handlers to the middleware and let it decide based on the JSON-RPC method and the contract address
+	router.Handle("/", PostRpcRequestMiddleware(rpcProxyHandler, erc721UniversalMintingHandler, storage)).Methods("POST")
```
zorjak: Is it possible to pass ` h handleInterface`  as an arg to the function `PostRpcRequestMiddleware`  and inside of it define (`rpcProxyHandler` and `erc721UnversalMintingHandler`)

cmd/server/api/handler.go
```diff
@@ -20,12 +20,15 @@ func (h *HTTPClientWrapper) Do(req *http.Request) (*http.Response, error) {
 }
 
 type HandlerInterface interface {
-	PostRPCHandler(w http.ResponseWriter, r *http.Request)
+	PostRPCProxyHandler(w http.ResponseWriter, r *http.Request)
+	UniversalMintingRPCHandler(w http.ResponseWriter, r *http.Request)
+	SetJsonRPCRequest(req JSONRPCRequest)
 }
 
 type Handler struct {
```
zorjak: Usually, if I define in the same package interface and struct that implements that interface I make the interface public and struct the same name but private. 

in this case, it would be 

`type Handler interface`

`type hanlder struct` .

Does it make sense to have the same approach here?

zorjak (changes requested):

cmd/server/api/rpcmiddleware.go
```diff
@@ -0,0 +1,144 @@
...
+
+func handleEthCallMethod(w http.ResponseWriter, r *http.Request, req JSONRPCRequest, standardHandler, erc721Handler http.Handler, st scan.Storage) {
+	var params ParamsRPCRequest
+	if len(req.Params) == 0 || json.Unmarshal(req.Params[0], &params) != nil {
+		reportError(w, "Error parsing params or missing params", http.StatusBadRequest)
+		return
+	}
+
+	// Check for universal minting method.
+	remoteMinting, err := isUniversalMintingMethod(params.Data)
```
AlexRipoll: I'd suggest prefixing remoteMinting with `is` since it's a boolean, and you're already following this pattern with isInContractList a few lines below.

cmd/server/api/rpcmiddleware.go
```diff
@@ -0,0 +1,144 @@
...
+	var req JSONRPCRequest
+	if err := json.Unmarshal(body, &req); err != nil {
+		reportError(w, "Error parsing JSON request", http.StatusBadRequest)
+		slog.Error("error parsing JSON request", "err", err)
+		return
+	}
+	if req.JSONRPC != "2.0" {
+		http.Error(w, "Invalid JSON-RPC version", http.StatusBadRequest)
+		return
+	}
```
AlexRipoll: Shouldn't this code reside in the `validateJSONRPCPostRequest` function, considering it's related to validation?

cmd/server/api/rpcmiddleware_test.go
```diff
@@ -0,0 +1,187 @@
...
+			contentType:        "application/json",
+			method:             "POST",
+			expectedStatusCode: http.StatusBadRequest,
+			expectedResponse:   "Invalid JSON-RPC version\n",
+			handlerToBeCalled:  "none",
+		},
+	}
+
+	// Run tests
+	for _, ttest := range tests {
```
AlexRipoll: Kindly rename it using `tt` and `tt := tt` so that we maintain consistency throughout the entire project.

internal/platform/rpc/erc721/calldata.go
```diff
@@ -0,0 +1,111 @@
...
+
+// getInputArgs returns a map of input arguments extracted from the calldata.
+func (b CallData) getInputArgs() (map[string]interface{}, error) {
+	sig, err := b.methodSignature()
+	if err != nil {
+		return nil, err
+	}
+
+	argdata := b[ShortAddressLength:]
+	if len(argdata)%32 != 0 {
```
AlexRipoll: Consider adding a constant for the integer 32 to enhance clarity

AlexRipoll (changes requested):

AlexRipoll (approved):

zorjak (approved):

PR #23, opened 2023-11-14
feature/UN-66 integrate database as storage

luispdm:


cmd/main.go
```diff
@@ -31,34 +36,88 @@ func run() error {
 	setLogger(c.Debug)
 	c.LogFields()
 
+	db, err := badger.Open(badger.DefaultOptions(c.Path).WithLoggingLevel(badger.ERROR))
```
zorjak, line 39: I think that we mentioned that we are going to add chainID in the path, right? I think that would be good to do.

cmd/main.go
```diff
@@ -67,16 +126,55 @@ func run() error {
...
+	if err != nil {
+		return fmt.Errorf("error getting chain ID from database: %w", err)
+	}
+
+	if chainIdDB == "" {
+		if err = repositoryService.SetChainID(chainId.String()); err != nil {
+			return fmt.Errorf("error setting chain ID in database: %w", err)
+		}
+	} else {
+		if chainId.String() != chainIdDB {
```
zorjak: Personally, I like more solutions where chainID is put in the storage path. 
The reason is this. We may want to run (and test) the node locally and we want to test it for many chains. If I understand this well, I would have to clean the db each time we switch ownership chain.

AlexRipoll: You can substitute `else` with an `else if chainId.String() != chainIdDB` to avoid an extra level of idententaion

cmd/main.go
```diff
@@ -96,31 +194,120 @@ func runScan(ctx context.Context, c *config.Config, client scan.EthClient, s sca
...
+			if err = repositoryService.SetEvoChainCurrentBlock(strconv.FormatUint(lastBlock+1, 10)); err != nil {
+				slog.Error("error occurred while storing current block", "err", err.Error())
+				break
+			}
 			startingBlock = lastBlock + 1
 		}
 	}
 }
 
+func getStartingBlock(ctx context.Context, startingBlockDB string, configStartingBlock uint64, client scan.EthClient) (uint64, error) {
```
AlexRipoll, line 288: Why is `startingBlockDB` of type string?

cmd/server/api/rpcmiddleware.go
```diff
@@ -116,14 +115,13 @@ func handleEthCallMethod(w http.ResponseWriter, r *http.Request, req *JSONRPCReq
 	}
 }
 
-func isContractInList(contractAddress string, st scan.Storage) (bool, error) {
-	list, err := st.ReadAll(context.Background())
+func isContractInList(contractAddress string, repositoryService repository.Service) (bool, error) {
+	list, err := repositoryService.GetAllERC721UniversalContracts()
```
zorjak, line 120: Why just not store contract in a db and check is it there. like storing contacts in a map.

 It would consume a bit more space but if we have many contracts and this function is called very often, I think this is not optimal.

internal/platform/storage/badger.go
```diff
@@ -0,0 +1,63 @@
...
+		returnValue, err = item.ValueCopy(returnValue)
+		if err != nil {
+			return err
+		}
+		return nil
+	})
+	return returnValue, err
+}
+
+func (b Badger) GetKeysWithPrefix(prefix []byte) ([][]byte, error) {
```
zorjak, line 43: Do we need this function? It does not look very fast in my opinion. There is some loop that does not seem optimal. Maybe I am wrong.

internal/scan/scan.go
```diff
@@ -14,17 +14,28 @@ import (
...
+	eventApprovalSigHash               = generateEventSignatureHash(eventApprovalName, "address", "address", "uint256")
+	eventApprovalForAllSigHash         = generateEventSignatureHash(eventApprovalForAllName, "address", "address", "bool")
+	eventNewERC721UniversalSigHash     = generateEventSignatureHash(eventNewERC721Universal, "address", "string")
+	eventNewCollectionSigHash          = generateEventSignatureHash(eventNewCollection, "uint64", "address")
+	eventMintedWithExternalURISigHash  = generateEventSignatureHash(eventMintedWithExternalURI, "uint64", "uint96", "address", "string", "uint256")
+	eventEvolvedWithExternalURISgiHash = generateEventSignatureHash(eventEvolvedWithExternalURI, "uint64", "uint256", "string")
+	EventNewCollectionSigHash          = generateEventSignatureHash(eventNewCollection, "uint64", "address")
+	EventMintedWithExternalURISigHash  = generateEventSignatureHash(eventMintedWithExternalURI, "uint64", "uint96", "address", "string", "uint256")
+	EventEvolvedWithExternalURISgiHash = generateEventSignatureHash(eventEvolvedWithExternalURI, "uint64", "uint256", "string")
```
AlexRipoll: Make it unexported

internal/scan/scan.go
```diff
@@ -211,37 +198,91 @@ func (s scanner) ScanEvents(ctx context.Context, fromBlock, toBlock *big.Int) ([
...
+			// Ownership events
+			case eventTransferSigHash:
+				transfer, err := parseTransfer(&eventLogs[i], &erc721UniversalAbi)
+				if err != nil {
+					if err != eventTopicsError {
+						return nil, err
+					}
+					slog.Warn("incorrect number of topics found in Transfer event",
+						"topics_found", len(eventLogs[i].Topics),
+						"topics_expected", 4)
```
AlexRipoll, line 224: Why is this check required? If an event matches with the hash signature, it should not be possible to have fewer topics than those specified in the event declaration.

AlexRipoll (changes requested):

internal/repository/repository.go
```diff
@@ -0,0 +1,110 @@
...
+func (s *Service) GetEvoChainCurrentBlock() (string, error) {
+	value, err := s.get(evoCurrentBlock)
+	if err != nil {
+		return "", err
+	}
+	return string(value), nil
+}
+
+func (s *Service) SetCurrentBlock(value string) error {
+	return s.storageService.Set([]byte(currentBlock), []byte(value))
```
zorjak, line 97: I thought that the current block is going to be per contract. what happened with that?

internal/repository/repository.go
```diff
@@ -0,0 +1,110 @@
+package repository
```
zorjak, line 1: Is it possible to rename this to something else? This naming `repository` is to general and looks like Java:):):)

why not `state`  as we talked before?:)

internal/repository/repository.go
```diff
@@ -0,0 +1,110 @@
...
+	storageService storage.Storage
+}
+
+func New(s storage.Storage) Service {
+	return Service{
+		s,
+	}
+}
+
+func (s *Service) StoreERC721UniversalContracts(universalContracts []model.ERC721UniversalContract) error {
```
zorjak, line 28: I don't understand this function. I thought that we are going to store contracts in DB as we discover them. 

so why slice and not just `model.ERC721UniversalContract`?

internal/scan/scan.go
luispdm: Unused, they must have been the remains of solving merge conflicts. Thanks

zorjak (approved):

cmd/server/api/rpcmiddleware.go
zorjak: maybe it is too much in this PR.

zorjak (changes requested):

cmd/main.go
luispdm: For sake of simplicity we store the block number as string in the storage. Storing it as `uint64` would mean dealing with `binary.LittleEndian` and `binary.BigEndian` and to be honest I have no idea whether we should use one or the other.
I am open to change it if you know what should we use

internal/scan/scan.go
luispdm: While scanning the whole chain we found some events with less topics, or, no topics at all. As a safeguard, instead of panicking, we chose to implement the check

cmd/main.go
luispdm: You still can test on various chains, but you have to tell the node to use a different directory with `-storage_path`.

We decided to go with this after a brief discussion with the product owner. I initially proposed your solution, but we agreed to implement it the other way (what we currently have), in case we wanna support future non-EVM chains that don't have the concept of a chain id.

However, while working on the solution, I noticed that the approach we were currently working on doesn't work with non-EVM chains either.

Either way, both approaches are fine. I'm totally open to accept your solution, but I'd kindly ask you to do it in another task.

cmd/main.go
luispdm: Read [here](https://github.com/freeverseio/laos-universal-node/pull/23#discussion_r1394338704)

internal/repository/repository.go
luispdm: You won 😆 We are tracking the block universally, one for the ownership chain and another for the evolution chain. Initial approach = keep it simple. We can optimize it in the future

internal/platform/storage/badger.go
kudlaj: this is the most optimal way to get the contracts because we're only fetching keys not values

internal/repository/repository.go
kudlaj: this was discarded (for now)

internal/repository/repository.go
kudlaj: This is not the state. The state is in the storage package! This is a repository function thats why we named it repository...

luispdm: I've asked chat gpt about names suggestions but I didn't like anything of what it proposed.

I'm open to rename the folder `data`, but I would leave this for a future task.
Why? Because repository will get bigger and bigger, as it will require more retrieval and store methods. We were thinking about specific repositories, each per data type, so to better separate responsibilities. Example.
chainid.go:
```go
type chainID struct {}

func (c *chainId) GetChainID() {
	// ...
}
```
currentblock.go:
```go
type currentBlock struct {}

func (c *currentBlock) GetCurrentBlock() {
	// ...
}
```
service.go:
```go
type Service struct {
	*chainId
	*contract
	*currentBlock
	// ....
}
```

All the files would still be stored in the data/repository folder.

AlexRipoll (approved):

cmd/server/api/rpcmiddleware.go
kudlaj: ok, we added a TODO and do the refactor in a different PR

internal/platform/storage/badger.go
luispdm: It is actually very fast, this is what dgraph (the company behind badger) claims: https://dgraph.io/docs/badger/get-started/#key-only-iteration and this is the benchmark they published in their blog https://dgraph.io/blog/images/badger-benchmarks.png (look at the bottom-right graph).
That picture is taken from this article: https://dgraph.io/blog/post/badger/

You can look at the benchmark data here: https://github.com/dgraph-io/badger-bench/blob/master/BENCH-rocks.txt

As the keys are stored in the LSM tree and not in the log files, key-only iteration is quite fast. As the DB is not searching for values, it doesn't have to scan log files.

P.S. I just found this other benchmark: https://dgraph.io/blog/post/badger-lmdb-boltdb/. I just hope they are not straight-out liars 😆

internal/repository/repository.go
luispdm: It is exactly what we are doing.

`main`, which calls this function, previously calls `ScanNewUniversalEvents`, which returns []model.ERC721UniversalContract`.

`StoreERC721UniversalContracts` performs a single `tx.Commit()` so the write to the DB happens once.

cmd/main.go
zorjak: you could forget to set -storage_path and you could end up in dirty DB. So I still think that chainID in path is a better solution. But, I agree, lets do that in another task.

cmd/server/api/rpcmiddleware.go
zorjak: ok

internal/platform/storage/badger.go
zorjak: ok

internal/repository/repository.go
zorjak: ok. I understand that for now we are going with that. I was not so strong with opinion with universal current block but I guess for now it is ok to keep it .

internal/repository/repository.go
zorjak: lets keep it as is and merge this because we are under pressure. But I am not 100% sure in this.

internal/repository/repository.go
zorjak: ok

zorjak (approved):

PR #48, opened 2023-11-28
dont process events until evo is not synced with ownership blocks that have been processed

zorjak:


cmd/main.go
```diff
@@ -331,6 +377,16 @@ func scanEvoChain(ctx context.Context, c *config.Config, client scan.EthClient,
...
+			headers, err := client.HeaderByNumber(ctx, big.NewInt(int64(nextStartingBlock)))
+			if err != nil {
+				slog.Error("error retrieving block headers", "err", err.Error())
+				break
+			}
+
+			if err = tx.SetCurrentEvoBlockTimestamp(headers.Time); err != nil {
+				slog.Error("error storing block headers", "err", err.Error())
+				break
+			}
```
AlexRipoll: I believe the code should be placed before the line `nextStartingBlock := lastScannedBlock.Uint64() + 1`, and we should store `lastScannedBlock` instead. However, I don't think this will impact the behavior.

AlexRipoll (approved):

kudlaj (approved):

luispdm (approved):

PR #53, opened 2023-12-04
cleaning some debug logs

zorjak:
optimize a bit of log writing. it was printing too many `no new universal contracts found` for every event when that was found. For me it was impossible to see anything useful in Grafana due to many of these logs.

### DIFF
--- a/internal/scan/scan.go
+++ b/internal/scan/scan.go
@@ -6,7 +6,6 @@ import (
 	"log/slog"
 	"math/big"
 	"regexp"
-	"strconv"
 	"strings"
 
 	"github.com/ethereum/go-ethereum"
@@ -205,7 +204,7 @@ func (s scanner) ScanNewUniversalEvents(ctx context.Context, fromBlock, toBlock 
 	}
 
 	if len(contracts) > 0 {
-		slog.Info("found " + strconv.Itoa(len(contracts)) + " new universal contracts")
+		slog.Info("universal contracts found", "contracts", len(contracts))
 	}
 
 	return contracts, nil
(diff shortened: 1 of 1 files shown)


internal/scan/scan.go
```diff
@@ -201,12 +201,13 @@ func (s scanner) ScanNewUniversalEvents(ctx context.Context, fromBlock, toBlock
...
 			}
 			contracts = append(contracts, c)
-
-		default:
-			slog.Debug("no new universal contracts found")
 		}
 	}
 
+	if len(contracts) > 0 {
+		slog.Info("found " + strconv.Itoa(len(contracts)) + " new universal contracts")
```
luispdm: Please change this to:
```go
slog.Info("universal contracts found", "contracts", len(contracts))
```
That is the intended way to log fields with `slog` (so you can get rid of `strconv` too)

internal/scan/scan.go
```diff
@@ -159,6 +160,7 @@ func NewScanner(client EthClient, contracts ...string) Scanner {
 
 // ScanEvents returns the ERC721 events between fromBlock and toBlock
 func (s scanner) ScanNewUniversalEvents(ctx context.Context, fromBlock, toBlock *big.Int) ([]model.ERC721UniversalContract, error) {
+	slog.Info("scanning universal events ", "fromBlock", fromBlock, "toBlock", toBlock)
```
luispdm, line 162: Can you do the same thing for this line here https://github.com/freeverseio/laos-universal-node/blob/fix/remove-some-logs/internal/scan/scan.go#L249 and move it at the top of `ScanEvents`, please? Thank you

luispdm (changes requested):

internal/scan/scan.go
zorjak: changed

internal/scan/scan.go
zorjak: added

luispdm (approved):

kudlaj (approved):

PR #55, opened 2023-12-05
optimizing evo event processing

zorjak:
only events that are not processed are considered when processing a chunk of blocks


cmd/main.go
```diff
@@ -627,18 +627,17 @@ func discoverContracts(ctx context.Context, client scan.EthClient, s scan.Scanne
...
-	var ownershipContractEvoBlock uint64
-	for i := range mintedEvents {
-		if mintedEvents[i].Timestamp > timestampContract {
-			break
+	for index, mintedEvent := range mintedEvents {
+		if mintedEvent.Timestamp > timestampContract {
+			return uint64(index), nil
 		}
-		if err := tx.Mint(contract, &mintedEvents[i]); err != nil {
+		localMintedEvent := mintedEvent // Create a local copy of local minted event to solve linter issue
```
luispdm: I propose to do a for range on the index as we did here -> https://github.com/freeverseio/laos-universal-node/pull/57/commits/8503a69ac770f48ddcb2f3535a247a440f7287f8

internal/state/state.go
```diff
@@ -65,8 +65,8 @@ type EvolutionContractState interface {
 }
 
 type OwnershipBlockState interface {
-	SetCurrentEvoBlockForOwnershipContract(contract string, blockNumber uint64) error
-	GetCurrentEvoBlockForOwnershipContract(contract string) (uint64, error)
+	SetCurrentEvoEventIndexForOwnershipContract(contract string, blockNumber uint64) error
+	GetCurrentEvoEventIndexForOwnershipContract(contract string) (uint64, error)
```
luispdm: I propose renaming these with `Events` instead of `Event`, so to understand we are tracking the index of a slice

internal/state/block/ownership/block.go
```diff
@@ -22,11 +22,11 @@ func NewService(tx storage.Tx) *service {
 	}
 }
 
-func (s *service) SetCurrentEvoBlockForOwnershipContract(contract string, number uint64) error {
+func (s *service) SetCurrentEvoEventIndexForOwnershipContract(contract string, number uint64) error {
```
luispdm: I just realized the key prefix we use is still `ownership_contract_evo_current_block_`, while it should be something like: `ownership_contract_evo_events_current_index_`.

I would also suggest moving `SetCurrentEvoEventIndexForOwnershipContract` and `GetCurrentEvoEventIndexForOwnershipContract` to another package, as the current one deals with blocks. Maybe the new package name could be "event"? Just a suggestion, I am open to other proposals

cmd/server/api/universalmintingrpchandler.go
```diff
@@ -41,6 +41,7 @@ func (h *UniversalMintingRPCHandler) HandleUniversalMinting(jsonRPCRequest JSONR
 	} else if !exists {
 		return getErrorResponse(fmt.Errorf("method not supported"), jsonRPCRequest.ID)
 	} else {
+		slog.Debug("UniversalMintingMethod call", "params", params.Data)
```
luispdm: We don't need this. Remove it, please

kudlaj (approved):

cmd/main.go
```diff
@@ -727,46 +727,38 @@ func updateStateWithTransfer(contract string, tx state.Tx, modelTransferEvent *m
...
-	// example:
-	// 2 events on block 10, you store the first, updatedBlock == 10. a ctrl + c comes, you don't store the second event and the method returns
-	// can it be that the transaction is committed?
+func updateStateWithMint(ctx context.Context, client scan.EthClient, contract string, tx state.Tx, mintedEvent *model.MintedWithExternalURI) error {
 	lastTaggedBlock, err := tx.GetLastTaggedBlock(common.HexToAddress(contract))
 	if err != nil {
-		return 0, err
+		return err
 	}
+	slog.Debug("updateStateWithMint", "lastTaggedBlock", lastTaggedBlock, "contract", contract)
```
luispdm: We don't need this, can you remove it, please? Same for line 710 and 551

cmd/main.go
```diff
@@ -652,43 +651,42 @@ func loadMerkleTree(tx state.Tx, contractAddress common.Address) error {
...
 			if err := updateStateWithTransfer(contract, tx, &modelTransferEvents[transferIndex]); err != nil {
 				return 0, err
 			}
 			transferIndex++
 		// all transfer events have been processed => process remaining minted events
 		case transferIndex >= len(modelTransferEvents):
-			block, err := updateStateWithMint(ctx, client, contract, tx, &mintedEvents[mintedIndex], ownershipContractEvoBlock, lastBlock)
-			if err != nil {
-				return 0, err
+			if mintedEvents[ownershipContractEvoEventIndex].Timestamp < lastBlockTimestamp {
```
luispdm, line 671: This should be an if-else

cmd/main.go
```diff
@@ -727,46 +727,38 @@ func updateStateWithTransfer(contract string, tx state.Tx, modelTransferEvent *m
...
 
 	for {
 		blockToTag := lastTaggedBlock + 1
 		timestamp, err := getTimestampForBlockNumber(ctx, client, uint64(blockToTag))
 		if err != nil {
-			return 0, err
+			return err
 		}
-		if uint64(blockToTag) > lastBlock || timestamp > mintedEvent.Timestamp {
+		if timestamp > mintedEvent.Timestamp {
```
luispdm: This should be >=

luispdm (changes requested):

cmd/main.go
zorjak: ok. that solved linter issue. thanks

internal/state/state.go
zorjak: ok

internal/state/block/ownership/block.go
zorjak: I've changed the prefix name. 

I would not create another package. I would rather name the `block` package into `synchronization` or something like that because its purpose is to deal with that stuff, right? what do you think?

cmd/server/api/universalmintingrpchandler.go
zorjak: ok

cmd/main.go
zorjak: ok

cmd/main.go
zorjak: ok

cmd/main.go
zorjak: ok

cmd/main.go
luispdm: Please change it to:
```go
for i := range mintedEvents
// ...
```
We don't need `mintedEvent` just for returning the error, you can use `mintedEvents[i]`

zorjak: changed

internal/state/block/ownership/block.go
luispdm: Ok. Either "sync" or "synchronization", as you wish

cmd/main.go
luispdm: Please, do it for lines 710 and 551 as well. Thanks.
551:
`slog.Debug("tagRootsUntilBlock", "lastTaggedBlock", lastTaggedBlock, "contractAddress", contractsAddress[i])`
710:
`slog.Debug("updateStateTransfer", "lastTaggedBlock", lastTaggedBlock, "contract", contract)`

zorjak: ok

internal/state/block/ownership/block.go
zorjak: changed to sync

internal/state/state.go
```diff
@@ -65,8 +65,8 @@ type EvolutionContractState interface {
 }
 
 type OwnershipBlockState interface {
```
luispdm: Rename these interfaces to `OwnershipSyncState` and `ÈvolutionSyncState` please

luispdm (approved):

internal/state/state.go
zorjak: changed

PR #56, opened 2023-12-11
Feature/un 96 adding ability for batch rpc calls

kudlaj:


cmd/server/api/rpcproxyhandler.go
```diff
@@ -2,46 +2,41 @@ package api
...
-		return
+		return getErrorResponse(fmt.Errorf("error creating request: %w", err), rpcId)
 	}
 
 	// Forward headers the request
-	proxyReq.Header = r.Header
+	for name, values := range r.Header {
+		for _, value := range values {
+			// we don't want to forward the Accept-Encoding header because we don't want to receive a gzipped response
+			if name != "Accept-Encoding" {
```
zorjak, line 29: should we explicitly check that the value is `gzip`? Or do we just always ignore this header? whenever encoding is there?

zorjak (approved):

cmd/server/api/rpcproxyhandler.go
kudlaj: if we don't put it in the headers, it will never be returned as gziped. So we don't have to check.

kudlaj: I is not necessary to check, we just ignore it because we don't want to have a response gziped

luispdm (approved):

PR #57, opened 2023-12-12
implement tokenURI functionality

luispdm:


cmd/main.go
```diff
@@ -628,11 +628,12 @@ func discoverContracts(ctx context.Context, client scan.EthClient, s scan.Scanne
 
 func updateStateWithMintEvents(contract common.Address, tx state.Tx, mintedEvents []model.MintedWithExternalURI, timestampContract uint64) (uint64, error) {
 	var ownershipContractEvoBlock uint64
-	for _, mintedEvent := range mintedEvents {
+	for _, mEvent := range mintedEvents {
+		mintedEvent := mEvent
```
kudlaj: This is redundant. Why not use directly mEvent? mEvent is already a copy of the element in mintedEvents, so, no need to make a copy of a copy.

zorjak (approved):

cmd/main.go
luispdm: The linter was raising the following concern: "Implicit memory aliasing in for loop" because I was reusing the same loop variable.
In these case I can either reassign the variable or do a `for range`. I accepted your change request and removed the other variable, we now have a `for range`.

kudlaj (approved):

PR #61, opened 2023-12-19
Feature/un 101 Translate blocktags for proxy calls

kudlaj:
- If a query is triggered with “latest” block, the “unode” will process it in the latest block it has processed, not the latest block on the ownershipchain.

- If a query is triggered with a block number larger than the latest processes block by the uNode it fails in the same way that a regular node will fail if you ask for a block not processed yet.


cmd/server/api/rpcmethods.go
```diff
@@ -0,0 +1,353 @@
...
+}
+
+var rpcMethodsWithHash = map[string]RPCMethod{
+	"eth_getBlockByHash":                    RPCMethodEthGetBlockByHash,
+	"eth_getTransactionReceipt":             RPCMethodEthGetTransactionReceipt,
+	"eth_getTransactionByHash":              RPCMethodEthGetTransactionByHash,
+	"eth_getTransactionByBlockHashAndIndex": RPCMethodEthGetTransactionByBlockHashAndIndex,
+}
+
+//	var rpcMethodsWithCountByHash = map[string]RPCMethod{
```
zorjak, line 119: should we remove this comment if it is not needed? or at least write a comment on why this is commented (not supported)

zorjak (approved):

cmd/server/api/rpcmethods.go
```diff
@@ -0,0 +1,355 @@
...
+	ReplaceBlockTag(req *JSONRPCRequest, method RPCMethod, blockNumberUnode string) (*JSONRPCRequest, error)
+}
+
+type ProxyRPCMethodManager struct{}
+
+func NewProxyRPCMethodManager() RPCMethodManager {
+	return &ProxyRPCMethodManager{}
+}
+
+func (b *ProxyRPCMethodManager) HasRPCMethodWithBlocknumber(methodName string) (RPCMethod, bool) {
```
AlexRipoll: Camel case typo in `HasRPCMethodWithBlocknumber`

AlexRipoll (changes requested):

cmd/server/api/rpcmiddleware.go
```diff
@@ -15,7 +15,7 @@ type JSONRPCRequest struct {
 	ID      *json.RawMessage  `json:"id,omitempty"`
 }
 
-type ParamsRPCRequest struct {
+type EthCallParamsRPCRequest struct {
```
luispdm: If you do not plan to use it in external packages tests, this should be unexported

cmd/server/api/rpcmethods.go
```diff
@@ -0,0 +1,355 @@
...
+
+import (
+	"encoding/json"
+	"fmt"
+	"math/big"
+)
+
+type RPCMethod int
+
+type BlockTag string
```
luispdm: Make it unexported, as well as: FilterObject, Block, Transaction and the constants Latest, Pending etc. and CompareHex(), please

cmd/server/api/rpcproxyhandler.go
```diff
@@ -49,6 +64,19 @@ func (h *RPCProxyHandler) HandleProxyRPC(r *http.Request, req JSONRPCRequest) RP
 	if err != nil {
 		return getErrorResponse(fmt.Errorf("error getting JSON RPC response: %w", err), req.ID)
 	}
+	// check if we have to replace the block tag
```
luispdm: I think this comment refers to linees 16-26 instead of 68-78

cmd/server/api/rpcmethods.go
```diff
@@ -0,0 +1,333 @@
...
+	case *Block:
+		return value.Number, nil
+	case *Transaction:
+		return value.BlockNumber, nil
+	}
+
+	return "", fmt.Errorf("unknown type for unmarshalling")
+}
+
+func replaceBlockTagWithHash(req *JSONRPCRequest, position int, blockNumberHash string) error {
```
luispdm: Rename "blockNumberHash" to "blockNumberUnode" here, as well as in "replaceBlockTagFromObject" and "getBlockNumber", please.

I also think that this function should be called: "replaceBlockTagWithBlockNumber".

luispdm (changes requested):

AlexRipoll (approved):

luispdm (approved):

PR #62, opened 2023-12-20
increase DB mem table size to 1GB

zorjak:


cmd/main.go
```diff
@@ -57,7 +57,7 @@ func run() error {
 	setLogger(c.Debug)
 	c.LogFields()
 
-	db, err := badger.Open(badger.DefaultOptions(c.Path).WithLoggingLevel(badger.ERROR))
+	db, err := badger.Open(badger.DefaultOptions(c.Path).WithLoggingLevel(badger.ERROR).WithMemTableSize(1<<30))
```
luispdm, line 64: Can you add the following comment above, please?
```go
// "WithMemTableSize" increases MemTableSize to 1GB (1<<30 is 1GB). This increases the transaction size to about 153MB (15% of MemTableSize)
```

luispdm (changes requested):

cmd/main.go
zorjak: ok

luispdm (approved):

AlexRipoll (approved):

PR #64, opened 2023-12-21
return err when it's supposed to

luispdm:


kudlaj (approved):

zorjak (approved):

cmd/main.go
```diff
@@ -314,7 +316,7 @@ func processUniversalBlockRange(ctx context.Context, c *config.Config, client sc
 
 	if err = tx.Commit(); err != nil {
 		slog.Error("error committing transaction", "err", err.Error())
-		return nil
+		return err
```
zorjak, line 320: maybe these error (similar with other errors) can be wrapped with the text in slog. And then you return error from this function you log it.

zorjak (approved):

PR #65, opened 2023-12-27
extract evo scanning and processing events function from main to worker folder

zorjak:


internal/core/worker/evolution/worker.go
```diff
@@ -0,0 +1,299 @@
...
+	"time"
+
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/freeverseio/laos-universal-node/internal/config"
+	"github.com/freeverseio/laos-universal-node/internal/platform/model"
+	"github.com/freeverseio/laos-universal-node/internal/scan"
+	"github.com/freeverseio/laos-universal-node/internal/state"
+)
+
+type ReorgError struct {
```
luispdm: Make it unexported

internal/core/worker/evolution/worker.go
```diff
@@ -0,0 +1,299 @@
...
+			slog.Info("context canceled")
+			return nil
+		default:
+			lastBlock, err := getLastBlock(ctx, w.client, startingBlock, w.blocksRange, w.blocksMargin)
+			if err != nil {
+				slog.Error(err.Error())
+				break
+			}
+            if lastBlock < startingBlock {
+				slog.Debug("scanEvochain, last calculated block is behind starting block, waiting...",
```
luispdm: Change `scanEvochain` to `evolution worker`

internal/core/worker/evolution/worker.go
```diff
@@ -0,0 +1,299 @@
...
+	block       uint64
+	chainHash   common.Hash
+	storageHash common.Hash
+}
+
+func (e ReorgError) Error() string {
+	return "reorg error"
+}
+
+type Worker interface {
```
luispdm, line 28: This interface isn't used anywhere

internal/core/worker/evolution/worker_test.go
```diff
@@ -0,0 +1,596 @@
+package worker
```
luispdm, line 1: As we test the private function `storeMintedWithExternalURIEventsByContract`, I'd rename the file `worker_wb_test.go`, the convention we have adopted so far to test private stuff (wb stands for white box)

luispdm (changes requested):

internal/core/worker/evolution/worker.go
zorjak: ok

internal/core/worker/evolution/worker.go
zorjak: it is used used in main.go. when we run this worker in go rutine.

internal/core/worker/evolution/worker.go
zorjak: ok

internal/core/worker/evolution/worker_test.go
zorjak: if you are agree lets keep it as is. I am still thinking about moving things in process package and doing small refact. when I move tests there I will have clearer where this function should stay and where the test should be

luispdm: As we're in the midst of a refactor, I'm ok with this for now

internal/core/worker/evolution/worker.go
luispdm: It is not, as `NewWorker` returns the private struct `worker`

zorjak: ups. you are right. 

Fixed. NewWorker returns the interface now.

PR #66, opened 2023-12-27
scan only ownership contracts pointing to the targeted evochain

luispdm:


internal/config/config.go
```diff
@@ -80,3 +95,18 @@ func getDefaultStoragePath() string {
 	}
 	return path.Join(homeDir, ".universalnode")
 }
+
+func setGlobalConsensusAndParachain(c *Config) error {
+	switch {
+	case strings.Contains(strings.ToLower(c.EvoRpc), "caladan"):
+		c.GlobalConsensus = caladanGlobalConsensus
+		c.Parachain = caladanParachain
+	case strings.Contains(strings.ToLower(c.EvoRpc), "klaos"):
```
zorjak: Maybe I am wrong but I think this is a bug. evoRpc is a domain name. 
It can be that evoRpc provider runs `caladan` node but gives it domain name that contains `klaos`?

Maybe we should call some node API to decide is it `caladan` or `klaos`? maybe by calling ChainId?

zorjak (changes requested):

internal/config/config.go
luispdm: Agreed. Changes pushed

zorjak (approved):

PR #68, opened 2023-12-28
fix nil errors in handler

luispdm:
- `ReplaceBlockTag` does not return an additional `*JSONRPCRequest`, as it already changes the receiving parameter "req"
- no need to run `CheckBlockNumberFromResponseFromHashCalls` when `response.Result` is `nil` (`CheckBlockNumberFromResponseFromHashCalls` also checks if `response.Result` is `nil` as a first instruction)
- do not include `result` in RPCResponse when there's an error
- include `result: null` when the underlying L1 node returns it


internal/core/worker/evolution/worker.go
```diff
@@ -68,7 +68,7 @@ func (w *worker) Run(ctx context.Context) error {
 				slog.Error(err.Error())
 				break
 			}
-            if lastBlock < startingBlock {
+			if lastBlock < startingBlock {
```
zorjak, line 71: can you remove these changes from this PR?

internal/core/worker/evolution/worker_test.go
```diff
@@ -225,7 +225,7 @@ func TestScanEvoChainOnce(t *testing.T) {
 			client, scanner := getMocks(t)
 			state, tx2 := getMocksFromState(t)
 
-            evoWorker := NewWorker(&tt.c, client, scanner, state)
+			evoWorker := NewWorker(&tt.c, client, scanner, state)
```
zorjak, line 228: similar as for worker

zorjak (changes requested):

internal/core/worker/evolution/worker.go
luispdm: As agreed in private, this change will stay in the PR

internal/core/worker/evolution/worker_test.go
luispdm: As agreed in private, this change will stay in the PR

zorjak (approved):

PR #69, opened 2024-01-01
un-122-extract scan universal functions to worker

zorjak:


internal/core/worker/ownership/worker.go
```diff
@@ -0,0 +1,698 @@
...
+		var existingContracts []string
+		existingContracts, err = tx.GetExistingERC721UniversalContracts(contracts)
+		if err != nil {
+			slog.Error("error occurred checking if user-provided contracts exist in storage", "err", err.Error())
+			return err
+		}
+		contractsAddress = append(contractsAddress, existingContracts...)
+	} else {
+		dbContracts := tx.GetAllERC721UniversalContracts()
+		contractsAddress = append(contractsAddress, dbContracts...)
```
luispdm: This can be simplified to: `contractsAddress = existingContracts` in the `if` and `contractsAddress = tx.GetAllERC721UniversalContracts()` in the `else` branch

internal/core/worker/ownership/worker.go
```diff
@@ -0,0 +1,698 @@
...
+				transferIndex++
+			}
+		}
+	}
+}
+
+func updateStateWithTransfer(contract string, tx state.Tx, modelTransferEvent *model.ERC721Transfer) error {
+	// TODO if transfer event's timestamp is ahead of global evo chain current block's timestamp => wait X seconds and read again the global evo chain current block from DB
+	// we must wait because there might have been mint events whose timestamp is < this transfer event
+	// maybe it is worth storing the global evo chain current block's timestamp also?!
```
luispdm: I think we can get rid of this worry TODO as we currently store the evochain block timestamp

luispdm (changes requested):

internal/core/worker/ownership/worker.go
zorjak: ok

internal/core/worker/ownership/worker.go
zorjak: ok

internal/core/worker/ownership/worker.go
zorjak: ok

luispdm (approved):

luispdm (approved):

PR #70, opened 2024-01-02
un-123 move universal process to separate module

zorjak:
- extract universal contract processor from worker
- extract  the discoverer to a separate package
- extract  the validator to a separate pacakge


internal/platform/state/sync/ownership/sync.go
```diff
@@ -25,48 +27,52 @@ func NewService(tx storage.Tx) *service {
...
-func (s *service) SetCurrentOwnershipBlock(number uint64) error {
-	return s.tx.Set([]byte(currentBlock), []byte(strconv.FormatUint(number, 10)))
+	var block model.Block
+
+	decoder := gob.NewDecoder(bytes.NewBuffer(value))
+	if err := decoder.Decode(&block); err != nil {
+		return defaultBlock, err
+	}
+	return block, nil
 }
```
luispdm: The content of these two functions is identical to the ones in `sync/evolution`. I suggest moving them to `sync/sync.go`, whether as package functions or struct methods

internal/core/processor/universal/discoverer/discoverer.go
```diff
@@ -0,0 +1,150 @@
...
+}
+
+type discoverer struct {
+	client    scan.EthClient
+	contracts []string
+	scanner   scan.Scanner
+	validator uValidator.Validator
+}
+
+func NewDiscoverer(
```
luispdm: I suggest renaming this "New", to avoid package name repetition when calling the method externally

internal/core/processor/universal/discoverer/validator/validator.go
```diff
@@ -0,0 +1,58 @@
...
+type Validator interface {
+	Validate(event scan.EventNewERC721Universal) (model.ERC721UniversalContract, error)
+}
+
+type validator struct {
+	globalConsensus string
+	parachain       uint64
+}
+
+func NewValidator(globalConsensus string, parachain uint64) Validator {
```
luispdm: I suggest renaming this "New", to avoid package name repetition when calling the method externally

internal/core/processor/universal/updater/updater.go
```diff
@@ -0,0 +1,248 @@
...
+		lastBlockData model.Block,
+	) error
+}
+
+type updater struct {
+	client  scan.EthClient
+	scanner scan.Scanner
+}
+
+func NewUpdater(client scan.EthClient, scanner scan.Scanner) Updater {
```
luispdm: I suggest renaming this "New", to avoid package name repetition when calling the method externally

internal/core/worker/evolution/worker.go
```diff
@@ -48,6 +48,7 @@ func (w *worker) Run(ctx context.Context) error {
 		default:
 			lastBlock, err := executeEvoBlockRange(ctx, w, startingBlock)
 			if err != nil {
+				slog.Error(err.Error())
```
luispdm: As we commented on your previous PR, let's remove this instruction as well as the one at line 71, as almost all the errors are logged in the called functions already.

The only error that needs logging is the final instruction of `updateLastBlockData`, at `internal/core/processor/evolution/processor.go`. I suggest replacing that instruction for the following:
```go
if err = tx.SetLastEvoBlock(block); err != nil {
	slog.Error("error occurred while setting LaosEvolution end range block",
		"lastBlock", lastBlock, "err", err.Error())
	return err
}
return nil
```

internal/core/worker/evolution/worker_wb_test.go
```diff
@@ -117,6 +116,7 @@ func TestExecuteBlockRange(t *testing.T) {
 		startingBlock := uint64(100)
 
 		ctx := context.TODO()
+		t.Helper()
```
luispdm: This is a test, not a helper function, remove this line, please

internal/core/worker/evolution/worker_wb_test.go
```diff
@@ -47,7 +47,7 @@ func TestExecuteBlockRange(t *testing.T) {
 		lastBlock, err := executeEvoBlockRange(ctx, worker, startingBlock)
 		assertError(t, nil, err)
 		if lastBlock != 99 {
-			t.Fatalf(`got last block %d, expected 99`, lastBlock)
+			t.Fatalf(`got last block %d, expected 90`, lastBlock)
```
luispdm: I remember commenting this while reviewing 125: the message should print `expected 99`. Is it me not understanding the test or did you have issues while rebasing?

internal/core/processor/universal/discoverer/validator/validator.go
```diff
@@ -0,0 +1,58 @@
...
+	return &validator{
+		globalConsensus: globalConsensus,
+		parachain:       parachain,
+	}
+}
+
+func (v *validator) Validate(event scan.EventNewERC721Universal) (model.ERC721UniversalContract, error) {
+	contractGlobalConsensus, err := event.GlobalConsensus()
+	if err != nil {
+		slog.Warn("error parsing collection address for contract", "contract", event.NewContractAddress,
```
luispdm: "global consensus" instead of "collection address"

internal/core/processor/universal/discoverer/validator/validator.go
```diff
@@ -0,0 +1,58 @@
...
+			"base_uri", event.BaseURI)
+		return model.ERC721UniversalContract{}, err
+	}
+	contractParachain, err := event.Parachain()
+	if err != nil {
+		slog.Warn("error parsing collection address for contract", "contract", event.NewContractAddress,
```
luispdm: "parachain" instead of "collection address"

internal/core/processor/universal/processor.go
```diff
@@ -0,0 +1,228 @@
...
+
+	tx := p.stateService.NewTransaction()
+	defer tx.Discard()
+
+	lastEvoBLockData, err := tx.GetLastEvoBlock()
+	if err != nil {
+		return false, err
+	}
+
+	slog.Debug("IsEvoSyncedWithOwnership", "evo_bock_number", lastEvoBLockData.Number, "evo_block_timestamp", lastEvoBLockData.Timestamp,
```
luispdm: Change "evo_bock_number" to "evo_block_number"

internal/core/processor/universal/processor.go
```diff
@@ -0,0 +1,228 @@
...
+	configBlocksRange   uint64
+	configBlocksMargin  uint64
+	discoverer          contractDiscoverer.Discoverer
+	updater             contractUpdater.Updater
+}
+
+func NewProcessor(client scan.EthClient,
+	stateService state.Service,
+	scanner scan.Scanner,
+	configStartingBlock,
```
luispdm: I won't insist here but I suggest replacing configStartingBlock, configBlocksMargin, configBlocksRange, and contracts for config.Config, so to reduce the number of parameters.

Up to you if you want to pass config.Config or *config.Config

internal/core/processor/universal/processor.go
```diff
@@ -0,0 +1,228 @@
...
+	}
+
+	return true, nil
+}
+
+func (p *processor) ProcessUniversalBlockRange(ctx context.Context, startingBlock, lastBlock uint64) error {
+	tx := p.stateService.NewTransaction()
+	defer tx.Discard()
+
+	lastBlockData, err := updateLastBlockData(ctx, tx, p.client, lastBlock)
```
luispdm: I have a proposal.
I suggest changing the name of `updateLastBlockData` to `getLastBlockData`. The function will just return a `model.Block` corresponding to `lastBlock`, without setting the data in the DB.
Then we call `tx.SetLastOwnershipBlock` right before `tx.Commit`.

I know that if there's an error the transaction is discarded anyways, but logically it makes more sense in my head to set the block in the DB only when everything succeedes

internal/core/processor/universal/updater/updater.go
```diff
@@ -0,0 +1,248 @@
...
+}
+
+func getFirstOwnershipBlockAfterMintEvent(ctx context.Context,
+	client scan.EthClient,
+	contract string,
+	tx state.Tx,
+	mintedEvent *model.MintedWithExternalURI,
+) (uint64, error) {
+	// TODO at this point this function uses GetLastTaggedBlock() to find the first ownership block before the mint event
+	// I suggest in the future implementation that in the separate goroutine crate mapping between ownership block
```
luispdm: I don't understand.
Can you rephrase this, please?

internal/core/processor/universal/updater/updater.go
```diff
@@ -0,0 +1,248 @@
...
+
+type Updater interface {
+	GetModelTransferEvents(
+		ctx context.Context,
+		startingBlock,
+		lastBlock uint64,
+		contracts []string,
+	) (map[string][]model.ERC721Transfer, error)
+
+	UpdateStates(
```
luispdm: `UpdateState` seems more appropriate to me, we only have one state

luispdm (changes requested):

internal/core/processor/universal/discoverer/discoverer.go
zorjak: ok

internal/core/processor/universal/discoverer/validator/validator.go
zorjak: ok

internal/core/processor/universal/discoverer/validator/validator.go
zorjak: ok

internal/core/processor/universal/discoverer/validator/validator.go
zorjak: ok

internal/core/processor/universal/processor.go
zorjak: ok. yes I agree with you that updating makes more sense to be in the end. It is how it is done in evochain actually. here I had to do at the beggining since I will need last block. but we can put this update in the end.

internal/core/processor/universal/processor.go
zorjak: ok

internal/core/processor/universal/updater/updater.go
zorjak: ok

internal/core/processor/universal/updater/updater.go
zorjak: ok

internal/core/processor/universal/updater/updater.go
zorjak: ok

internal/core/worker/evolution/worker.go
zorjak: ok . added line

internal/core/worker/evolution/worker_wb_test.go
zorjak: It should be 99. this when there is nil and we return startingBlock-1 because in worker we do startingBlock=lastBlock+1

internal/core/worker/evolution/worker_wb_test.go
zorjak: ok

internal/platform/state/sync/ownership/sync.go
zorjak: I did not understand how to do that. Also ind also that the parameter `latBlock` is different for ownership and evolution.

internal/core/processor/universal/processor.go
zorjak: ok

internal/core/worker/evolution/worker.go
luispdm: Thanks. You have to remove `slog.Error(err.Error())` at line 71 only

internal/core/processor/universal/processor.go
luispdm: Sorry Zoran, I should have explained myself better. I would still keep configStartingBlock, configBlocksMargin, configBlocksRange, and contracts as processor's fields, I would only change the signature of `NewProcessor` with *config.Config. Having *config.Config as a processor field can be dangerous as we might accidentally modify the original values.

So I would do:
```go
type processor struct {
	client              scan.EthClient
	stateService        state.Service
	scanner             scan.Scanner
	configStartingBlock uint64
	configBlocksMargin  uint64
	configBlocksRange   uint64
	contracts           []string
	discoverer          contractDiscoverer.Discoverer
	updater             contractUpdater.Updater
}

func NewProcessor(client scan.EthClient,
	stateService state.Service,
	scanner scan.Scanner,
	c *config.Config,
	discoverer contractDiscoverer.Discoverer,
	updater contractUpdater.Updater,
) *processor {
	return &processor{
		client:              client,
		stateService:        stateService,
		scanner:             scanner,
		configStartingBlock: c.StartingBlock,
		configBlocksMargin:  uint64(c.BlocksMargin),
		configBlocksRange:   uint64(c.BlocksRange),
		contracts:           c.Contracts,
		discoverer:          discoverer,
		updater:             updater,
	}
}
```

internal/core/processor/universal/processor.go
luispdm: Thanks.
You should also move
```go
slog.Debug("setting ownership end range block hash for block number",
		"blockNumber", block.Number(), "blockHash", block.Hash(), "parentHash", block.ParentHash())
```
right before `tx.SetLastOwnershipBlock`

internal/platform/state/sync/ownership/sync.go
luispdm: My idea:

`sync/sync.go`:
```go
package sync

import (
	"bytes"
	"encoding/gob"

	"github.com/ethereum/go-ethereum/common"
	"github.com/freeverseio/laos-universal-node/internal/platform/model"
	"github.com/freeverseio/laos-universal-node/internal/platform/storage"
)

func SetLastBlock(tx storage.Tx, key string, block model.Block) error {
	var buf bytes.Buffer
	encoder := gob.NewEncoder(&buf)

	if err := encoder.Encode(block); err != nil {
		return err
	}

	return tx.Set([]byte(key), buf.Bytes())
}

func GetLastBlock(tx storage.Tx, key string) (model.Block, error) {
	defaultBlock := model.Block{
		Number:    0,
		Timestamp: 0,
		Hash:      common.Hash{},
	}

	value, err := tx.Get([]byte(key))
	if err != nil {
		return defaultBlock, err
	}
	if value == nil {
		return defaultBlock, nil
	}

	var block model.Block

	decoder := gob.NewDecoder(bytes.NewBuffer(value))
	if err := decoder.Decode(&block); err != nil {
		return defaultBlock, err
	}
	return block, nil
}
```

evolution/sync.go:
```go
package evolution

import (
	"github.com/freeverseio/laos-universal-node/internal/platform/model"
	"github.com/freeverseio/laos-universal-node/internal/platform/state/sync"
	"github.com/freeverseio/laos-universal-node/internal/platform/storage"
)

const (
	lastBlock = "evo_last_block"
)

type service struct {
	tx storage.Tx
}

func NewService(tx storage.Tx) *service {
	return &service{
		tx: tx,
	}
}

func (s *service) SetLastEvoBlock(block model.Block) error {
	return sync.SetLastBlock(s.tx, lastBlock, block)
}

func (s *service) GetLastEvoBlock() (model.Block, error) {
	return sync.GetLastBlock(s.tx, lastBlock)
}
```

And similarly, `ownership/sync.go`

internal/core/worker/universal/worker.go
```diff
@@ -0,0 +1,139 @@
...
+
+func (w *worker) executeUniversalBlockRange(ctx context.Context,
+	evoSynced bool,
+	startingBlock,
+	lastBlock uint64,
+) (previousLastBlock uint64, wasEvoSynced bool, err error) {
+	if evoSynced {
+		lastBlock, err = w.processor.GetLastBlock(ctx, startingBlock)
+		if err != nil {
+			slog.Debug("Error obtaining the last block", "error", err)
```
luispdm: We don't need this as the error is logged by `GetLastBlock`

internal/platform/state/sync/ownership/sync.go
zorjak: ok

internal/core/worker/evolution/worker.go
zorjak: removed everywhere:)

internal/core/processor/universal/processor.go
zorjak: ok

internal/core/processor/universal/processor.go
zorjak: ok

internal/core/worker/universal/worker.go
zorjak: ok

internal/platform/state/sync/ownership/sync.go
zorjak: ok

internal/core/processor/universal/discoverer/validator/validator_test.go
```diff
@@ -0,0 +1,92 @@
...
+	"github.com/freeverseio/laos-universal-node/internal/platform/model"
+	"github.com/freeverseio/laos-universal-node/internal/platform/scan"
+)
+
+func TestValidate(t *testing.T) {
+	t.Parallel()
+	tests := []struct {
+		name             string
+		globalConsensus  string
+		parachain        uint64
```
luispdm: Unused parameters

internal/core/processor/universal/discoverer/discoverer_test.go
```diff
@@ -0,0 +1,305 @@
...
+			startingBlock := uint64(100)
+			lastBlock := uint64(200)
+
+			d := cDiscoverer.New(client, []string{}, scanner, validator)
+
+			// Mock the scanner's ScanNewUniversalEvents method
+			scanner.EXPECT().ScanNewUniversalEvents(ctx, big.NewInt(int64(startingBlock)), big.NewInt(int64(lastBlock))).
+				Return([]scan.EventNewERC721Universal{tt.event}, tt.errorOnScanning)
+
+			if tt.errorOnScanning == nil {
```
luispdm: This tests has too many ifs. We will struggle quite a lot to fix the tests if we need to refactor discoverer.
I suggest splitting the tests between error test cases and successful test cases. At least we will have less conditionals.

I'm ok even without table-tests and just `t.Run` subtests with repeated code

internal/core/processor/universal/updater/updater_test.go
```diff
@@ -0,0 +1,203 @@
...
+	uUpdater "github.com/freeverseio/laos-universal-node/internal/core/processor/universal/updater"
+	mockScan "github.com/freeverseio/laos-universal-node/internal/platform/scan/mock"
+	mockTx "github.com/freeverseio/laos-universal-node/internal/platform/state/mock"
+	"github.com/freeverseio/laos-universal-node/internal/platform/state/tree/enumerated"
+	"github.com/freeverseio/laos-universal-node/internal/platform/state/tree/enumeratedtotal"
+	"github.com/freeverseio/laos-universal-node/internal/platform/state/tree/ownership"
+	"go.uber.org/mock/gomock"
+)
+
+func TestUpdateState(t *testing.T) {
```
luispdm, line 22: We need error test cases here, this function is important. If you don't have time today I can try to do it next week or we'll tackle this with a task

luispdm (changes requested):

internal/core/processor/universal/updater/updater_test.go
zorjak: yes, this is just a happy path for updater. I think this is going to change very soon so I did not insisted much writting tests. I could write some wb tests in updater but I think updater with the fixes we plan for optmization is going to be changed very soon. so I did not bother adding tests now to be honest.

internal/core/processor/universal/discoverer/validator/validator_test.go
zorjak: ok removed

internal/core/processor/universal/discoverer/discoverer_test.go
zorjak: ok

cmd/main.go
```diff
@@ -119,7 +122,11 @@ func run() error {
 	// Ownership chain scanner
 	group.Go(func() error {
 		s := scan.NewScanner(ownershipChainClient, c.Contracts...)
-		uWorker := worker.NewWorker(c, ownershipChainClient, s, stateService)
+
+		discoveryValidator := validator.New(c.GlobalConsensus, c.Parachain)
+		discoverer := contractDiscoverer.New(ownershipChainClient, c.Contracts, s, discoveryValidator)
+		updater := contractUpdater.New(ownershipChainClient, s)
+		uWorker := universalWorker.NewWorker(c, ownershipChainClient, s, stateService, discoverer, updater)
```
kudlaj: we should agree on packet.New or packet.NewSomething and do it always the same
now we have it all mixed up

internal/core/processor/evolution/processor.go
```diff
@@ -169,7 +169,14 @@ func updateLastBlockData(ctx context.Context, tx state.Tx, client scan.EthClient
 		Hash:      lastBlockData.Hash(),
 	}
 
-	return tx.SetLastEvoBlock(block)
+	err = tx.SetLastEvoBlock(block)
+	if err != nil {
+		slog.Error("error occurred while setting LaosEvolution end range block",
```
kudlaj: this error message is confusing, can you make it clearer?

internal/core/processor/universal/updater/updater_test.go
zorjak: actually maybe the happy path is enough. the rest would be just mocking calls that don't have any logic between

luispdm (approved):

luispdm (approved):

internal/core/processor/universal/updater/updater.go
```diff
@@ -0,0 +1,248 @@
...
+	return tx.Mint(common.HexToAddress(contract), mintedEvent)
+}
+
+func getFirstOwnershipBlockAfterMintEvent(ctx context.Context,
+	client scan.EthClient,
+	contract string,
+	tx state.Tx,
+	mintedEvent *model.MintedWithExternalURI,
+) (uint64, error) {
+	// TODO at this point this function uses GetLastTaggedBlock(). I think the first ownership block after mint event
```
kudlaj: why do we keep this TODO?

internal/core/worker/universal/worker.go
```diff
@@ -0,0 +1,130 @@
...
+type Worker interface {
+	Run(ctx context.Context) error
+}
+
+type worker struct {
+	waitingTime time.Duration
+	processor   universal.Processor
+}
+
+func NewWorker(c *config.Config,
```
kudlaj: why NewWorker and not New?

internal/core/worker/universal/worker.go
```diff
@@ -0,0 +1,130 @@
...
+		select {
+		case <-ctx.Done():
+			slog.Info("context canceled")
+			return nil
+		default:
+			slog.Debug("executing block range", "startingBlock", startingBlock, "lastBlock", lastBlock, "evoSynced", evoSynced)
+			prevLastBlock, wasEvoSynced, err := w.executeUniversalBlockRange(ctx, evoSynced, startingBlock, lastBlock)
+			if err != nil {
+				var reorgErr universal.ReorgError
+				if errors.As(err, &reorgErr) {
```
kudlaj, line 66: if the error is not a reorgErr it is not logged

kudlaj (approved):

internal/core/processor/universal/updater/updater.go
zorjak: I will remove TODO part. I think this will going to be changed. probably when you do reorg handling.

for sure there we will have to have mapping evoBlock-ownershipBlock. Then this function will be just a call to that mapping.

cmd/main.go
zorjak: I will fix this

internal/core/processor/evolution/processor.go
zorjak: ok

internal/core/worker/universal/worker.go
zorjak: @luispdm I will return log of this error. I know that you say we log errors when they appear but it is possible always that we miss some and we just return error. better have the logs at the top also.

internal/core/worker/universal/worker.go
zorjak: New will be

internal/core/processor/evolution/processor.go
zorjak: fixed

internal/core/worker/universal/worker.go
luispdm: > if the error is not a reorgErr it is not logged

it is logged everywhere by `executeUniversalBlockRange` and the functions of the `processor`

PR #73, opened 2024-01-05
Feature/un 95 Adding CORS abilities

kudlaj:


zorjak (approved):

cmd/server/api/router_test.go
```diff
@@ -0,0 +1,69 @@
...
+
+	"github.com/freeverseio/laos-universal-node/cmd/server/api"
+	"github.com/freeverseio/laos-universal-node/cmd/server/api/mock"
+	stateMock "github.com/freeverseio/laos-universal-node/internal/platform/state/mock"
+	"github.com/gorilla/mux"
+	"go.uber.org/mock/gomock"
+)
+
+func TestCORS(t *testing.T) {
+	mockCtrl := gomock.NewController(t)
```
luispdm: Add `t.Parallel()`.

If you plan on doing the typical table test (`for _, tc := range tests { tc.Run(...)`), don't forget to add `t.Parallel` inside the subtest too.

Also if you plan on doing that, add the property `name` to the `tests` struct

cmd/server/api/router_test.go
```diff
@@ -0,0 +1,69 @@
...
+	api.Routes(mockRPCHandler, router, state)
+
+	ts := httptest.NewServer(router)
+	defer ts.Close()
+
+	client := ts.Client()
+
+	tests := []struct {
+		method       string
+		url          string
```
luispdm: No need for this parameter, you can hardcode "/" when creating a request

cmd/server/api/router_test.go
```diff
@@ -0,0 +1,69 @@
...
+	}
+
+	for _, tc := range tests {
+		req, err := http.NewRequest(tc.method, ts.URL+tc.url, http.NoBody)
+		if err != nil {
+			t.Errorf("could not create request: %v", err)
+		}
+		res, err := client.Do(req)
+		if err != nil {
+			t.Errorf("could not send request: %v", err)
```
luispdm: If `res` is `nil`, the next check will panic.

Either you replace all `t.Errorf` calls for `t.Fatalf` or you check that `res` is not `nil` on every `if`

luispdm (changes requested):

cmd/server/api/router_test.go
kudlaj: I will leave this, maybe we want to test a url at some point

luispdm (approved):

PR #75, opened 2024-01-08
Feature/un 107

AlexRipoll:
- Filtering events by topics
- Removing Approval, ApproveForAll and NewCollection events from scanner

PR #77, opened 2024-01-09
UN-127

AlexRipoll:
moving EthClient interface from scan to platform/blockchain pkg


luispdm:
closing PR as the change was integrated in #75

PR #81, opened 2024-01-12
Recover from Reorg: Feature/un 72

kudlaj:
With this feature we add the functionality to recover from a reorg in the ownership chain


cmd/main.go
```diff
@@ -119,6 +119,28 @@ func run() error {
 		}
 	})
 
+	// Ownership delete old block tags
+	group.Go(func() error {
```
zorjak, line 125: do we need to have this at all? Is there any other reason that this exist besides to clean database. To me this looks like the same as deleting tag root for older blocks. we don't save too much size in DB but we consume the bandwidth of it?

internal/core/processor/universal/processor.go
```diff
@@ -81,24 +87,128 @@ func (p *processor) VerifyChainConsistency(ctx context.Context, startingBlock ui
...
+		}
+		if errCommit := contractTx.Commit(); errCommit != nil {
+			return nil, errCommit // Handle commit error
+		}
+	}
+	// set last ownership block to block without reorg
+	if err := tx.SetLastOwnershipBlock(*blockWithoutReorg); err != nil {
+		return nil, err
+	}
+	// deleting all blockhashes after the block without reorg
```
zorjak: can you change this to `block hashes`? remove space

internal/core/processor/universal/processor.go
```diff
@@ -81,24 +87,128 @@ func (p *processor) VerifyChainConsistency(ctx context.Context, startingBlock ui
...
 	}
+	return p.checkBlockForReorg(ctx, lastBlockDB)
+}
+
+func (p *processor) RecoverFromReorg(ctx context.Context, currentBlock uint64) (*model.Block, error) {
+	// Start a transaction
+	tx := p.stateService.NewTransaction()
+	defer tx.Discard()
+
+	storedBlockNumbers, err := tx.GetAllStoredBlockNumbers()
```
zorjak, line 107: I think from this line  `storedBlockNumbers, err := tx.GetAllStoredBlockNumbers()` until line 118 code should be placed in one function. For instance and that function can be named `getValidBlock` (returns block for which reorg did not happen). 

I think this code:

```go
nextBlockNumberToCheck, err := getNextLowerBlockNumber(currentBlock, storedBlockNumbers)
	if err != nil { // no lower block number found
		// we get a safe block number to start from
		blockWithoutReorg = getSafeBlock(currentBlock)
    }
```
should go inside `p.CheckForReorgRecursive`. I see that function also calls `getSafeBlock` and `getNextLowerBlockNumber`.

Not mandatory from my side but I recomend to give it a try.

internal/core/processor/universal/processor.go
```diff
@@ -193,3 +304,32 @@ func getLastBlockData(ctx context.Context, client blockchain.EthClient, lastBloc
...
+	if err != nil {
+		slog.Error("error occurred checking out merkle tree at block number", "block_number", blockNumber,
+			"contract_address", contractAddress, "err", err)
+		return err
+	}
+
+	return nil
+}
+
+func getSafeBlock(currentBlockNumber uint64) *model.Block {
```
zorjak, line 323: I think this function could be simplified like this

```go
	if currentBlockNumber < safeBlockMargin {
		return &model.Block{ Number: 0}
    }
	return &model.Block{ Number: currentBlockNumber - safeBlockMargin}
```

internal/platform/state/sync/ownership/sync.go
```diff
@@ -24,12 +27,29 @@ func NewService(tx storage.Tx) *service {
 	}
 }
 
+func (s *service) SetOwnershipBlock(blockNumber uint64, block model.Block) error {
+	formatedOwnerhipBlockNumber := formatBlockNumber(blockNumber, blockNumberDigits)
```
zorjak: typo, put `formatedOwnershipBlockNumber` instad of `formatedOwnerhipBlockNumber`

internal/platform/state/sync/ownership/sync.go
```diff
@@ -24,12 +27,29 @@ func NewService(tx storage.Tx) *service {
...
+	return sync.SetBlock(s.tx, lastBlock, block)
 }
 
 func (s *service) GetLastOwnershipBlock() (model.Block, error) {
-	return sync.GetLastBlock(s.tx, lastBlock)
+	return sync.GetBlock(s.tx, lastBlock)
+}
+
+func (s *service) GetOwnershipBlock(blockNumber uint64) (model.Block, error) {
+	formatedOwnerhipBlockNumber := formatBlockNumber(blockNumber, blockNumberDigits)
```
zorjak: typo. put `formatedOwnershipBlockNumber`

internal/platform/state/state.go
```diff
@@ -62,6 +62,11 @@ type OwnershipSyncState interface {
 
 	SetLastOwnershipBlock(block model.Block) error
 	GetLastOwnershipBlock() (model.Block, error)
+	GetOwnershipBlock(blockNumber uint64) (model.Block, error)
+	SetOwnershipBlock(blockNumber uint64, block model.Block) error
+	GetAllStoredBlockNumbers() ([]uint64, error)
+	CleanStoredBlockNumbers() error
```
zorjak: I think that `CleanStoredBlockNumbers` and `DeleteStoredBlockNumbersNewerThanBlockNumber` in the essence do the same job. Maybe we should have only one functions which name would be 

`DeleteBlocksForBlockRange(startBlock, endBlock uint64)`

cmd/main.go
kudlaj: Yes we need this, otherwise we need to add a pagination when we retrieve the stored blocknumbers

internal/platform/state/state.go
kudlaj: not quite because CleanStoredBlockNumbers deletes all blocks that are not in the range from actual - 256  and DeleteStoredBlockNumbersNewerThanBlockNumber deletes all from blocknumber to newest.

internal/core/processor/universal/processor.go
```diff
@@ -81,24 +87,128 @@ func (p *processor) VerifyChainConsistency(ctx context.Context, startingBlock ui
...
+			return getSafeBlock(blockNumberToCheck), nil
+		}
+		return p.checkForReorgRecursive(ctx, tx, nextBlockToCheck, storedBlockNumbers)
+	default:
+		// Other error occurred
+		return nil, err
+	}
+}
+
+func getNextLowerBlockNumber(currentBlock uint64, storedBlockNumbers []uint64) (uint64, error) {
```
AlexRipoll: I recommend returning a boolean value instead of an error, as the absence of a lower block number should not be considered an error. In my opinion, this modification will enhance the clarity of the code at line 108.

internal/core/processor/universal/processor.go
```diff
@@ -81,24 +87,128 @@ func (p *processor) VerifyChainConsistency(ctx context.Context, startingBlock ui
...
 	}
-
 	// During the initial iteration, no hash is stored in the database, so this code block is bypassed.
 	if (lastBlockDB.Hash == common.Hash{}) {
 		return nil
 	}
+	return p.checkBlockForReorg(ctx, lastBlockDB)
+}
+
+func (p *processor) RecoverFromReorg(ctx context.Context, currentBlock uint64) (*model.Block, error) {
```
AlexRipoll, line 102: I recommend including a function description to enhance readability and facilitate understanding.

internal/platform/state/state.go
```diff
@@ -62,6 +62,11 @@ type OwnershipSyncState interface {
 
 	SetLastOwnershipBlock(block model.Block) error
 	GetLastOwnershipBlock() (model.Block, error)
+	GetOwnershipBlock(blockNumber uint64) (model.Block, error)
+	SetOwnershipBlock(blockNumber uint64, block model.Block) error
+	GetAllStoredBlockNumbers() ([]uint64, error)
+	DeleteOldStoredBlockNumbers() error
+	DeleteStoredBlockNumbersNewerThanBlockNumber(blockNumberRef uint64) error
```
AlexRipoll: Personally, I would name this function something like `DeleteOrphanBlockNumbers` instead

internal/core/worker/universal/worker.go
```diff
@@ -25,14 +25,24 @@ type worker struct {
...
 
+type Option func(*worker)
+
+// WithProcessor sets a custom processor for the worker.
+func WithProcessor(p universal.Processor) Option {
+	return func(w *worker) {
+		w.processor = p
+	}
+}
+
```
AlexRipoll: Why do we need this?

cmd/main.go
zorjak: ok

internal/core/processor/universal/processor.go
zorjak: can you rechecked this. as it is implemented now I think you have a bug. 

previously you were returning 0 in case that currentBlock is less then blockMargin. and now you return currentBlockNumber.

AlexRipoll (changes requested):

internal/core/worker/universal/worker.go
kudlaj: for the tests

internal/platform/state/state.go
zorjak: one more vote for this name.

zorjak (approved):

internal/core/worker/universal/worker.go
kudlaj: removed

AlexRipoll (approved):

PR #82, opened 2024-01-12
Feature/un 150

AlexRipoll:
Adding logic to only process evo blocks that have reached finality


internal/core/processor/evolution/client.go
```diff
@@ -0,0 +1,141 @@
...
+type HTTPClient interface {
+	Do(req *http.Request) (*http.Response, error)
+}
+
+type LaosHTTP struct {
+	Client HTTPClient
+	Url    string
+}
+
+func NewLaosHTTP(url string) LaosRPCRequests {
```
kudlaj: you should consider dependency injection here for HTTPClient (we use functional option pattern for this)

internal/core/processor/evolution/client.go
```diff
@@ -0,0 +1,141 @@
...
+			slog.Error("error closing response body", "err", err)
+		}
+	}()
+
+	if resp.StatusCode >= http.StatusBadRequest {
+		return fmt.Errorf("error in request to %s, got status code: %d", l.Url, resp.StatusCode)
+	}
+
+	if err := json.NewDecoder(resp.Body).Decode(result); err != nil {
+		return fmt.Errorf("error decoding JSON: %w", err)
```
kudlaj, line 95: maybe we should log the body here to get more info about the error

kudlaj (approved):

internal/core/processor/evolution/client.go
```diff
@@ -0,0 +1,141 @@
...
+	payload := []byte(fmt.Sprintf(`{"jsonrpc":"2.0","method":"chain_getBlock","params":[%q],"id":1}`, blockHash))
+
+	var response ChainGetBlock
+	if err := l.makeRequest(payload, &response); err != nil {
+		return nil, err
+	}
+
+	hexNumber := response.Result.Block.Header.Number
+
+	blockNumber, err := hexToDecimal(hexNumber)
```
zorjak, line 122: you are wrapping this error two times, here and in the function. Is this necessary?

internal/core/processor/evolution/client.go
```diff
@@ -0,0 +1,141 @@
...
+	return blockNumber, nil
+}
+
+func hexToDecimal(hex string) (*big.Int, error) {
+	// Remove the "0x" prefix
+	if len(hex) > 2 && hex[:2] == "0x" {
+		hex = hex[2:]
+	}
+
+	decimal, ok := new(big.Int).SetString(hex, 16)
```
zorjak, line 136: why you just don't do return new(big.Int).SetString(hex, 16)? if previous conditions are not meet it should return false on OK anyway.

internal/core/processor/evolution/client_wb_test.go
```diff
@@ -0,0 +1,44 @@
...
+	tests := []struct {
+		input     string
+		expected  *big.Int
+		expectErr bool
+	}{
+		{"1a", big.NewInt(26), false},
+		{"0x1a", big.NewInt(26), false},
+		{"abc", big.NewInt(2748), false},
+		{"0xabc", big.NewInt(2748), false},
+		{"invalid", nil, true},   // Invalid hex
```
zorjak, line 19: why ok is true if expected result is nil?

internal/core/processor/evolution/processor.go
```diff
@@ -98,6 +102,21 @@ func (p *processor) ProcessEvoBlockRange(ctx context.Context, startingBlock, las
...
+		if err != nil {
+			slog.Error("error occurred while checking latest finalized block", "err", err.Error())
+			return err
+		}
+		if ok {
+			break
+		}
+		slog.Debug("block not finalized, waiting for finality", "block", lastBlock)
+
+		time.Sleep(time.Millisecond * 500)
```
zorjak: this sleep and infinite loop potentialy won't allow you to turn off the node gracefully if you don't get last finalized block. you should use.

```go
func WaitBeforeNextScan(ctx context.Context, waitingTime time.Duration) {
	timer := time.NewTimer(waitingTime)
	select {
	case <-ctx.Done():
		timer.Stop()
	case <-timer.C:
	}
}
```


it is defined in worker.

zorjak (changes requested):
one potential smaller bug. two smaller suggestions.

internal/core/processor/evolution/client.go
AlexRipoll: It's a matter of personal preference and adhering to idiomatic conventions. This format appears more readable to me, at least.

internal/core/processor/evolution/client_wb_test.go
AlexRipoll: If the `hexToDecimal` functionis unable to convert to a big.Int, it will return both a nil value and an error. Could you please specify the exact aspect that is raising concerns for you?

internal/core/processor/evolution/processor.go
AlexRipoll: I'll be substituting it; however, even though this approach is an improvement, the existing method shouldn't lead to an infinite loop. We're requesting a block number, not a block hash, so eventually, a block with the specified number X will be generated.

zorjak (approved):

PR #83, opened 2024-01-16
Feature/un 151: Fixing reorg TX issues 

kudlaj:
In this feature if there is a reorg, we delete all root tags and set the last ownership block to the block without reorg before we do the checkout of the merkle trees. This way if we get an error checking out the trees for the different contracts we will not an inconsistent state.


zorjak (changes requested):
Please, increase coverage threshold to 75 here https://github.com/freeverseio/laos-universal-node/blob/main/.circleci/config.yml#L39

AlexRipoll (approved):

zorjak (approved):

PR #84, opened 2024-01-17
Refactor for better reorg detection

kudlaj:


internal/core/processor/universal/processor_test.go
```diff
@@ -167,121 +167,133 @@ func TestGetLastBlock(t *testing.T) {
...
+		name                       string
+		startingBlock              uint64
+		previousBlockHeader        *types.Header
+		previousBlockData          model.Block
+		blockHeader                *types.Header
+		blockData                  model.Block
+		discoverReturn             bool
+		updateReturn               map[string][]model.ERC721Transfer
+		expectedError              error
+		expectedTxCommit           int
```
zorjak: Can `commit `happen if non `nil` is expected? Maybe I am wrong, but I think it is not possible.

internal/core/processor/universal/processor_test.go
```diff
@@ -167,121 +167,133 @@ func TestGetLastBlock(t *testing.T) {
...
-		lastBlockDB            model.Block
-		lastBlockDBError       error
-		previousBlockData      *types.Header
-		previousBlockDataError error
-		expectedError          error
+		name                       string
+		startingBlock              uint64
+		previousBlockHeader        *types.Header
+		previousBlockData          model.Block
+		blockHeader                *types.Header
```
zorjak: I think BlockHeader and BlockData probably both contain the same information. Can you check if both of them are needed?

zorjak (approved):

internal/core/processor/universal/processor_test.go
kudlaj: yes, they are needed but the name is confusing. I will change it

internal/core/processor/universal/processor_test.go
kudlaj: you mean non nil error? Yes, but only if the error happens in the commit. But I prefer to have two different variables for clarity.

AlexRipoll (approved):

PR #85, opened 2024-01-18
replace SMT with jelyfish implementation from pokt network

zorjak:


zorjak:
closing this PR because we cannot use pokt network implementation since it deletes updated nodes so it does not preserve the history.

PR #89, opened 2024-01-23
optimization of enumeratedtotal trees

zorjak:
`TotalSupply` in `enumeratedTotal` is placed on index 0 of the tree and not in a separate key.


zorjak:
Unfortunately, the number of mints in transactions decreased. Probably as the number of assets increase we can see this benefit but for now don't merge this.

PR #90, opened 2024-01-23
Feature/un 160: Fixing problems with too many events in value

kudlaj:


internal/platform/state/contract/evolution/contract.go
```diff
@@ -24,29 +25,33 @@ func NewService(tx storage.Tx) *service {
...
 }
 
 func (s *service) GetMintedWithExternalURIEvents(contract string) ([]model.MintedWithExternalURI, error) {
-	value, err := s.tx.Get([]byte(eventsPrefix + strings.ToLower(contract)))
-	if err != nil {
-		return nil, err
-	}
-	if value == nil {
-		return nil, nil
+	events := s.tx.GetValuesWithPrefix([]byte(eventsPrefix + strings.ToLower(contract) + "_"))
```
zorjak, line 49: this line loads all the events. Probably it does not affect transaction size because it is just read, but at some point this is also going to be an issue (slow load) if the number of events becomes significantly big.  We should think about how to optimize this, also. Not sure if is it easy to fix. if not I would create jira task and leave it for the next sprint.

internal/platform/storage/badger/badger.go
```diff
@@ -156,3 +158,66 @@ func (t Tx) GetKeysWithPrefix(prefix []byte, reverse ...bool) [][]byte {
...
+		if err != nil {
+			continue
+		}
+		values = append(values, value)
+	}
+
+	return values
+}
+
+func (t Tx) FilterKeysWithPrefix(prefix []byte, from, to string) [][]byte {
```
zorjak, line 199: I don't see who uses this function?

internal/platform/storage/badger/badger.go
```diff
@@ -156,3 +158,66 @@ func (t Tx) GetKeysWithPrefix(prefix []byte, reverse ...bool) [][]byte {
...
+	opts := badger.DefaultIteratorOptions
+	opts.PrefetchValues = true
+	opts.Reverse = isReverse
+	opts.PrefetchSize = 100
+	iterator := t.tx.NewIterator(opts)
+	defer iterator.Close()
+
+	// Append 0xff to the prefix for the seek key if in reverse mode
+	var seekPrefix []byte
+	if isReverse {
```
zorjak, line 180: why do we need this reverse mode?

zorjak (changes requested):

internal/platform/state/contract/evolution/contract.go
```diff
@@ -24,29 +25,33 @@ func NewService(tx storage.Tx) *service {
...
-
-	if err := encoder.Encode(events); err != nil {
-		return err
+	for _, event := range events {
+		var buf bytes.Buffer
+		encoder := gob.NewEncoder(&buf)
+		if err := encoder.Encode(event); err != nil {
+			return err
+		}
+		if errSet := s.tx.Set([]byte(eventsPrefix+strings.ToLower(contract)+"_"+fmt.Sprint(event.BlockNumber)+"_"+fmt.Sprint(event.TxIndex)), buf.Bytes()); errSet != nil {
```
AlexRipoll: Maybe an approach like this one will make it easier to read. What do you think? 
`fmt.Sprintf("%s%s_%d_%d", eventsPrefix, strings.ToLower(contract), event.BlockNumber, event.TxIndex)`

AlexRipoll (approved):

internal/platform/storage/badger/badger.go
kudlaj: Nobody, but I leave it there because I think we will need it in the future

internal/platform/storage/badger/badger.go
kudlaj: it is not used right now, but we will need it eventually

internal/platform/state/contract/evolution/contract.go
kudlaj: jira task created: https://freeverse.atlassian.net/browse/UN-137

zorjak (approved):

PR #91, opened 2024-01-24
don't store array in leaf of enumerated tree. store each tokenId in a separate leaf

zorjak:
Optimization of enumerated Merkle tree. 

- don't store all the array of tokens in a leaf. store just one token per leaf, the same way as in enumerated total.
- this significantly increases the number of assets that we can store in a Merkle tree. 
- For transaction size of 150MB, and 10000 mints per block we were able to store up to 830000 assets can be store
- If the number of mints per block is smaller number increases even more.

without this implementation, we were able to store up to 30000 assets for the same conditions (more than 25x improvement)

### DIFF
--- a/internal/platform/state/v1/state_integration_test.go
+++ b/internal/platform/state/v1/state_integration_test.go
@@ -80,45 +80,66 @@ func TestLoadMerkleTreesWithBadger(t *testing.T) {
 			t.Fatalf(`got error "%s", expected "%s"`, err.Error(), expectedErr)
 		}
 	})
-	t.Run("successfully loads merkle trees in memory and mints 1400 assets", func(t *testing.T) {
+	t.Run("mint many assets in different transaction and then measure tx size", func(t *testing.T) {
 		// the test can run for 1400 assets and it fails for 1500 for this table size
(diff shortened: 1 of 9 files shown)


internal/platform/state/v1/state_integration_test.go
```diff
@@ -80,45 +80,68 @@ func TestLoadMerkleTreesWithBadger(t *testing.T) {
...
-		db := createBadger(t)
-		tx := createBadgerTransaction(t, db)
 		contract := common.HexToAddress("0x500")
+		db := createBadger(t)
 
-		err := tx.LoadMerkleTrees(contract)
-		if err != nil {
-			t.Errorf(`got error "%v" when no error was expected`, err)
-		}
+		fmt.Println("starting the test")
```
kudlaj: you could remove the Println ;)

kudlaj (approved):

AlexRipoll (approved):

internal/platform/state/v1/state_integration_test.go
zorjak: ok

PR #93, opened 2024-01-26
Feature/un 134

AlexRipoll:


internal/core/processor/universal/updater/updater.go
```diff
@@ -80,155 +88,171 @@ func (u *updater) UpdateState(
...
-		if err != nil {
-			return fmt.Errorf("error occurred retrieving evochain minted events for ownership contract %s and collection address %s: %w",
-				contract, collection.String(), err)
-		}
-
-		evoIndex, err := mergeAndUpdate(ctx, u.client, mintedEvents, modelTransferEvents[contract], contract, tx, lastBlockData.Timestamp)
-		if err != nil {
-			return fmt.Errorf("error updating state: %w", err)
-		}
+	slog.Debug("UpdateState", "startingBlock", startingBlock, "lastBlockData.Number", lastBlockData.Number)
```
kudlaj, line 96: Debug -> Error

internal/platform/state/contract/evolution/contract.go
```diff
@@ -27,31 +28,33 @@ func NewService(tx storage.Tx) *service {
...
+		strings.ToLower(contract),
+		formatNumberForSorting(blockNumber, blockNumberDigits))
+
+	events := s.tx.GetValuesWithPrefix([]byte(key))
 	var mintedEvents []model.MintedWithExternalURI
 	if len(events) == 0 {
 		return mintedEvents, nil
 	}
 
+	// TODO: this for loop probably can be avoided by decoding the whole slice of events
```
kudlaj: each event is individually encoded with the gob encoder and then stored under a unique key meaning each event is a separate entry in the storage system, so I don't think that this is possible. I would remove the TODO

internal/platform/state/contract/evolution/contract.go
```diff
@@ -27,31 +28,33 @@ func NewService(tx storage.Tx) *service {
...
+		formatNumberForSorting(event.BlockNumber, blockNumberDigits),
+		formatNumberForSorting(event.TxIndex, txIndexDigits))
+
+	return s.tx.Set([]byte(key), buf.Bytes())
 }
 
-func (s *service) GetMintedWithExternalURIEvents(contract string) ([]model.MintedWithExternalURI, error) {
-	events := s.tx.GetValuesWithPrefix([]byte(eventsPrefix + strings.ToLower(contract) + "_"))
+func (s *service) GetMintedWithExternalURIEvents(contract string, blockNumber uint64) ([]model.MintedWithExternalURI, error) {
+	slog.Debug("GetMintedWithExternalURIEvents", "contract", contract, "blockNumber", blockNumber)
```
kudlaj: I would remove this log

internal/platform/state/sync/evolution/sync.go
```diff
@@ -27,3 +33,49 @@ func (s *service) SetLastEvoBlock(block model.Block) error {
...
+
+// SetNextEvoEventBlockForOwnershipContract is used by evo processor for storing the next block that has events
+func (s *service) SetNextEvoEventBlock(contract string, blockNumber uint64) error {
+	value, err := s.tx.Get([]byte(lastEvoEventBlockPrefix + strings.ToLower(contract)))
+	if err != nil {
+		return err
+	}
+
+	uintValue := uint64(0)
+	if value == nil && len(value) > 0 {
```
kudlaj: this must to be wrong. Did you mean: if value != nil && len(value) > 0?
otherwise it never will enter the if

internal/core/processor/universal/updater/updater.go
```diff
@@ -80,155 +88,171 @@ func (u *updater) UpdateState(
...
+		return err
+	}
 
-		if err = tx.SetCurrentEvoEventsIndexForOwnershipContract(contract, evoIndex); err != nil {
-			return fmt.Errorf("error updating current evochain index %d for ownership contract %s: %w", evoIndex, contract, err)
+	for block := startingBlock; block <= lastBlockData.Number; block++ {
+		blockTime := blockTimestamps[block]
+		for _, contract := range contracts {
+			if blockWhenDiscovered, ok := newContracts[common.HexToAddress(contract)]; !ok {
+				if block < blockWhenDiscovered {
```
kudlaj, line 106: is this correct? It seems inverted. If it is correct please add some comments or make it a it more clearer to follow

internal/platform/state/v1/state.go
```diff
@@ -104,16 +116,50 @@ func (t *tx) setTreesForContract(
...
+
+	enumeratedTotalTree.SetRoot(accountData.EnumeratedTotalRoot)
+	enumeratedTotalTree.SetTotalSupply(accountData.TotalSupply)
+	enumeratedTree.SetRoot(accountData.EnumeratedRoot)
+	ownershipTree.SetRoot(accountData.OwnershipRoot)
+
+	return nil
+}
+
+// TODO check if it can be merged with LoadContractState
```
kudlaj: can we remove this TODO? or can we merge it with LoadContractState?

internal/platform/state/sync/evolution/sync_test.go
```diff
@@ -59,3 +65,50 @@ func TestSetGetLastEvoBlock(t *testing.T) {
...
+		if err != nil {
+			t.Errorf(`got error "%v" when no error was expected`, err)
+		}
+
+		err = tx.SetNextEvoEventBlock(common.HexToAddress("0x500").Hex(), 100)
+		if err != nil {
+			t.Errorf(`got error "%v" when no error was expected`, err)
+		}
+
+		block, err := tx.GetNextEvoEventBlock(common.HexToAddress("0x500").Hex(), 0)
```
kudlaj, line 84: add more cases. This will fail for any other block than 0 since there is a bug in the code, but the test does not detect it

internal/platform/state/sync/evolution/sync.go
```diff
@@ -27,3 +33,49 @@ func (s *service) SetLastEvoBlock(block model.Block) error {
...
 }
+
+// SetNextEvoEventBlockForOwnershipContract is used by evo processor for storing the next block that has events
+func (s *service) SetNextEvoEventBlock(contract string, blockNumber uint64) error {
+	value, err := s.tx.Get([]byte(lastEvoEventBlockPrefix + strings.ToLower(contract)))
+	if err != nil {
+		return err
+	}
+
+	uintValue := uint64(0)
```
kudlaj, line 44: can you change this name to nextNumber or something like that?

kudlaj (changes requested):

internal/platform/state/contract/evolution/contract.go
zorjak: ok

internal/platform/state/contract/evolution/contract.go
zorjak: I think you are right about this. I will remove the comment.

internal/core/processor/universal/updater/updater.go
zorjak: this basically mean that contract should be updated at the block on which it was discovered if it contains preminted events. if the block is less then the one when contract was discovered just skip it in this iteration

internal/platform/state/sync/evolution/sync.go
zorjak: I've put it nextValue

internal/platform/state/sync/evolution/sync.go
zorjak: this should be len(value)>0. That check is enough. I had simialr stuff on other place and just checking len is enough. It was bug that is fixed now

internal/platform/state/sync/evolution/sync.go
zorjak: this is actually just for converting `value` to uint. later it was again converted to the []byte. I have to do this check because if the value does not exist returns error, and I want it to be  0.

zorjak: considering that it is the same variable but just convert type I called it uintValue. I think that nextValue is not really a good option. Can we leave it as is?

internal/platform/state/sync/evolution/sync_test.go
zorjak: bug fixed and test added more cases. This was a good catch. Thank you.

internal/platform/state/v1/state.go
zorjak: yes. this was actually merged.

kudlaj (approved):

luispdm (approved):

internal/core/processor/universal/updater/updater.go
luispdm: `!ok` changed to `ok` to fix the bug, resolving convo

PR #101, opened 2024-02-02
increase resources in the chart with tolerations to schedule on more powerfull nodes

zorjak:


kudlaj (approved):

deployment/charts/universalnode/templates/statefulset.yaml
```diff
@@ -58,11 +58,16 @@ spec:
...
+          effect: "NoSchedule"
   volumeClaimTemplates:
   - metadata:
       name: storage
     spec:
       accessModes: [ "ReadWriteOnce" ]
       resources:
         requests:
-          storage: 100Gi
+          storage: 2000Gi
```
luispdm: Put `2Ti`, so it is exactly 2 terabytes

deployment/charts/universalnode/templates/statefulset.yaml
```diff
@@ -58,11 +58,16 @@ spec:
         volumeMounts:
         - name: storage
           mountPath: /app/.universalnode
+      tolerations:
+        - key: "pod"
+          operator: "Equal"
+          value: {{ .Values.tolerations }}
```
luispdm: You don't need to specify separate tolerations values, one is fine. If we merge your proposal we'd have to taint the node with 2 values. As per my understanding of the [kubernetes documentation](https://kubernetes.io/docs/concepts/scheduling-eviction/taint-and-toleration/), if the node is tainted with multiple key-values, for a pod to be deployed onto that node it must have all the key-values defined in its tolerations block

luispdm (changes requested):

deployment/charts/universalnode/templates/statefulset.yaml
zorjak: how much I understand this is going to be which node is tainted to what? maybe I am wrong. like one node (pod) is tainted to one and another with the second physical node?

deployment/charts/universalnode/templates/statefulset.yaml
zorjak: ok

deployment/charts/universalnode/templates/statefulset.yaml
zorjak: ok I think you are right. changed

luispdm (approved):

ccubu (approved):

PR #102, opened 2024-02-03
fix universal contract scanning error

zorjak:


internal/platform/scan/scan.go
```diff
@@ -191,12 +191,16 @@ func (s scanner) ScanNewUniversalEvents(ctx context.Context, fromBlock, toBlock
 		return nil, fmt.Errorf("error instantiating ABI: %w", err)
 	}
 
+	slog.Debug("event logs received", "logs", len(eventLogs))
```
kudlaj: I don't think this log is necessary

internal/platform/scan/scan.go
```diff
@@ -191,12 +191,16 @@ func (s scanner) ScanNewUniversalEvents(ctx context.Context, fromBlock, toBlock
...
 	contracts := make([]EventNewERC721Universal, 0)
 	for i := range eventLogs {
 		if len(eventLogs[i].Topics) == 0 {
 			continue
 		}
-
+		if len(eventLogs[i].Data) == 0 {
```
kudlaj, line 199: maybe we should move this to line 205 and print an WARN if it is an universal contract without data

internal/platform/scan/scan.go
luispdm: agree

luispdm: agree

internal/platform/scan/scan.go
luispdm: I don't think that is necessary, I would rather unify the if conditions:
```go
if len(eventLogs[i].Topics) > 0 && len(eventLogs[i].Data) > 0 && eventLogs[i].Topics[0].Hex() == eventNewERC721UniversalSigHash {
    newERC721Universal, err := parseNewERC721Universal(&eventLogs[i], &contractAbi)
    if err != nil {
        return nil, err
    }
    slog.Info("received event", eventNewERC721Universal, newERC721Universal)
    contracts = append(contracts, newERC721Universal)
}
```

kudlaj (approved):

internal/platform/scan/scan.go
zorjak: removed

internal/platform/scan/scan.go
zorjak: I prefere separate checks in validations. if you don't insist I would keep it as it is.

luispdm (approved):

PR #104, opened 2024-02-05
enviornment files for deployment

zorjak:


deployment/environments/dev/values.yaml
```diff
@@ -1,23 +1,19 @@
 networks:
   - name: polygon-mumbai
-    rpc: https://rpc-mumbai.maticvigil.com/
-    host: polygon-mumbai.unode.gorengine.com
+    host: polygon-mumbai-staging.unode.gorengine.com
     debug: true
-    startingBlock: "0"
+    startingBlock: 42400000
```
luispdm: I've looked for the ownership block of each ownership chain right after the first evo block of klaos.
These are the ones:
polygon mainnet -> **50442008**
polygon mumbai -> **42891324**
ethereum mainnet -> **18661200**

Please, replace the values accordingly in all the `.yaml` files

deployment/environments/prod/values.yaml
```diff
@@ -0,0 +1,28 @@
+networks:
+  - name: polygon-mainnet
+    host: polygon-mainnet.unode.klaos.laosfoundation.io/ 
+    debug: true
```
luispdm, line 4: I suggest removing the debug logs in prod. If you really want to keep them, I suggest removing them as soon as we publish v1.0

luispdm (changes requested):

deployment/environments/dev/values.yaml
zorjak: Thanks for these numbers. I will put these numbers as starting blocks.

deployment/environments/prod/values.yaml
kudlaj: +1 on that

kudlaj (approved):

deployment/environments/prod/values.yaml
zorjak: This also came to my mind. But since it is still not very stable I would keep debug in prod now, also. and yes, when we reach v1.0.0 we should remove debug logs. I agree.

luispdm (approved):

AlexRipoll (approved):

